PR #2, opened 2017-04-21
Fix ovirt-image-template flow

machacekondra:
This PR contains following:

 * Fixes the ovirt-image-template flow, so it always remove the VM, also in case when the template creation fails
 * Add a varibale to specify the template creation timeout
  * Changes the passwords.yml example to plaintext


machacekondra:
@mwperina

mwperina (commented):
Generally looks good, but please fix CI issue

examples/passwords.yml
```diff
@@ -1,7 +1,12 @@
 ---
-$ANSIBLE_VAULT;1.1;AES256
-39363136353036306238623430313265346233613033313163643434313662383365386532386533
-3639376638626338626535343939356664366638373837310a636438666330653133336231306338
-64353836306430363239396631363532663338326533313361303664386137393137326563626330
-6234333835393461640a346436633562396532363130353939646631623134303366393163313366
-33376638643565353338386538616331646665636237386131633965383865393633
+# As an example this file is keep in plaintext, if you want to
+# decrypt this file, please execute following command:
```
mykaul: decrypt -> encrypt

PR #3, opened 2017-05-15
ovirt-cluster-upgrade: add new role

machacekondra:
Add new cluster upgrade role.

### DIFF
--- a/roles/ovirt-cluster-upgrade/README.md
+++ b/roles/ovirt-cluster-upgrade/README.md
@@ -1,0 +1,60 @@
+oVirt cluster upgrade
+=========
+
+This role will iterate through all hosts in cluster and upgrades them.
+
+Requirements
+------------
+
+ * Ansible version 2.3 or higher
+ * Python SDK version 4 or higher
+
+Role Variables
+--------------
+
+| Name                    | Default value         |                                                     |
+|-------------------------|-----------------------|-----------------------------------------------------|
+| cluster_name            | Default               | Name of the cluster to be upgraded.                 |
+| stopped_vms             | UNDEF                 | List of VM names to be stopped before upgrade.      |
+| stop_pinned_to_host_vms | false                 | If true the pinned to host VMs will be stopped.     |
+| upgrade_timeout         | 1200                  | Timeout in seconds to wait for host to be upgraded. |
+| host_statuses           | [UP]                  | If the hosts are in any of the statuses specified by this list, the host will be upgraded. |
+| host_names              | [\*]                  | List of host names which should be upgraded.        |
+
+Dependencies
+------------
+
+No.
+
+Example Playbook
+----------------
+
+```yaml
+---
+- name: oVirt infra
+  hosts: localhost
+  connection: local
+  gather_facts: false
+
+  vars:
+    engine_url: https://ovirt-engine.example.com/ovirt-engine/api
+    engine_user: admin@internal
+    engine_password: 123456
+    engine_cafile: /etc/pki/ovirt-engine/ca.pem
+
+    cluster_name: production
+    stopped_vms:
+      - openshift-master-0
+      - openshift-node-0
+      - openshift-node-image
+
+  roles:
+    - ovirt-cluster-upgrade
+```
+
+[![asciicast](https://asciinema.org/a/122760.png)](https://asciinema.org/a/122760)
+
+License
+-------
+
+Apache License 2.0
--- a/roles/ovirt-cluster-upgrade/library/ovirt_clusters.py
+++ b/roles/ovirt-cluster-upgrade/library/ovirt_clusters.py
@@ -1,9 +1,23 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
 #
-# Copyright (c) 2017 Red Hat, Inc.
+# Copyright (c) 2016 Red Hat, Inc.
 #
-
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+#
 
 ANSIBLE_METADATA = {'metadata_version': '1.0',
                     'status': ['preview'],
@@ -159,8 +173,6 @@ options:
             - "C(legacy) - Legacy behavior of 3.6 version."
             - "C(minimal_downtime) - Virtual machines should not experience any significant downtime."
             - "C(suspend_workload) - Virtual machines may experience a more significant downtime."
-            - "C(post_copy) - Virtual machines should not experience any significant downtime.
-               If the VM migration is not converging for a long time, the migration will be switched to post-copy."
         choices: ['legacy', 'minimal_downtime', 'suspend_workload']
     serial_policy:
         description:
@@ -260,6 +272,7 @@ from ansible.module_utils.ovirt import (
     check_sdk,
     create_connection,
     equal,
+    get_link_name,
     ovirt_full_argument_spec,
     search_by_name,
 )
@@ -298,7 +311,6 @@ class ClustersModule(BaseModule):
         # legacy - 00000000-0000-0000-0000-000000000000
         # minimal downtime - 80554327-0569-496b-bdeb-fcbbf52b827b
         # suspend workload if needed - 80554327-0569-496b-bdeb-fcbbf52b827c
-        # post copy - a7aeedb2-8d66-4e51-bb22-32595027ce71
         migration_policy = self.param('migration_policy')
         if migration_policy == 'legacy':
             return '00000000-0000-0000-0000-000000000000'
@@ -306,8 +318,6 @@ class ClustersModule(BaseModule):
             return '80554327-0569-496b-bdeb-fcbbf52b827b'
         elif migration_policy == 'suspend_workload':
             return '80554327-0569-496b-bdeb-fcbbf52b827c'
-        elif migration_policy == 'post_copy':
-            return 'a7aeedb2-8d66-4e51-bb22-32595027ce71'
 
     def _get_sched_policy(self):
         sched_policy = None
@@ -436,7 +446,6 @@ class ClustersModule(BaseModule):
         )
 
     def update_check(self, entity):
-        sched_policy = self._get_sched_policy()
         migration_policy = getattr(entity.migration, 'policy', None)
         return (
             equal(self.param('comment'), entity.comment) and
@@ -465,7 +474,7 @@ class ClustersModule(BaseModule):
             equal(self.param('migration_compressed'), str(entity.migration.compressed)) and
             equal(self.param('serial_policy'), str(getattr(entity.serial_number, 'policy', None))) and
             equal(self.param('serial_policy_value'), getattr(entity.serial_number, 'value', None)) and
-            equal(self.param('scheduling_policy'), getattr(sched_policy, 'name', None)) and
+            equal(self.param('scheduling_policy'), get_link_name(self._connection, entity.scheduling_policy)) and
             equal(self._get_policy_id(), getattr(migration_policy, 'id', None)) and
             equal(self._get_memory_policy(), entity.memory_policy.over_commit.percent) and
             equal(self.__get_minor(self.param('compatibility_version')), self.__get_minor(entity.version)) and
@@ -512,10 +521,7 @@ def main():
         migration_bandwidth_limit=dict(default=None, type='int'),
         migration_auto_converge=dict(default=None, choices=['true', 'false', 'inherit']),
         migration_compressed=dict(default=None, choices=['true', 'false', 'inherit']),
-        migration_policy=dict(
-            default=None,
-            choices=['legacy', 'minimal_downtime', 'suspend_workload', 'post_copy']
-        ),
+        migration_policy=dict(default=None, choices=['legacy', 'minimal_downtime', 'suspend_workload']),
         serial_policy=dict(default=None, choices=['vm', 'host', 'custom']),
         serial_policy_value=dict(default=None),
         scheduling_policy=dict(default=None),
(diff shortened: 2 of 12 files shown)


machacekondra:
@mwperina

machacekondra:
@mgoldboi @oourfali @masayag

roles/ovirt-cluster-upgrade/library/ovirt_clusters.py
```diff
@@ -0,0 +1,567 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2016 Red Hat, Inc.
```
mykaul, line 4: 2017

roles/ovirt-cluster-upgrade/library/ovirt_clusters.py
```diff
@@ -0,0 +1,567 @@
...
+#
+# This file is part of Ansible
```
mykaul, line 6: Is it part of Ansible?

roles/ovirt-cluster-upgrade/tasks/pinned_vms.yml
```diff
@@ -0,0 +1,25 @@
...
+    - "{{ ovirt_vms | default([]) }}"
+  when: "{{ 'hosts' in item['placement_policy'] and item.placement_policy.hosts | length == 1 and item.status == 'up' }}"
+  register: host_ids_result
+
+- name: Create list of host IDs which has pinned VM
+  set_fact:
+    host_ids: "{{ host_ids_result.results | rejectattr('ansible_facts', 'undefined') | map(attribute='ansible_facts.host_ids_items') | list }}"
+
+- block:
+  - name: Shutdown pin to one host VMs
```
mykaul, line 16: I think this should be optional - same for HA VMs. If they are up, we may want to stop and let the user handle it beforehand.

mgoldboi:
@machacekondra, I think most of the items were captured. one thing that we might want to do is add couple of validations. probably the most needed is that the cluster is in up state (to avoid start upgrading on problematic conditions)with an option to bypass it, second item is hosts which their status isn't up treatment (whether to take those hosts in the upgrade flow or not)

oourfali:
@mgoldboi we don't have a status for clusters.
As for hosts that aren't up, maybe we can add a flag to enable/disable upgrading hosts that are already in maintenance. @mwperina - thoughts?

roles/ovirt-cluster-upgrade/README.md
```diff
@@ -0,0 +1,57 @@
...
+    engine_url: https://ovirt-engine.example.com/ovirt-engine/api
+    engine_user: admin@internal
+    engine_password: 123456
+    engine_cafile: /home/omachace/ovirt-engine41/etc/pki/ovirt-engine/ca.pem
+
+    cluster_name: production
+    low_prio_vms:
+      - openshift_master_0
+      - openshift_node_0
+      - openshift-node-image
```
mwperina: I'd prefer some non-product VM names, I think my-vm-1 and my-vm-2 are enough for this example

roles/ovirt-cluster-upgrade/library/ovirt_scheduling_policies_facts.py
```diff
@@ -0,0 +1,127 @@
+#!/usr/bin/python
```
mwperina, line 1: Please also add PR to Ansible for this new oVirt module

roles/ovirt-cluster-upgrade/tasks/main.yml
```diff
@@ -0,0 +1,128 @@
...
+      insecure: "{{ engine_insecure | default(true) }}"
+    when: ovirt_auth is undefined
+    register: loggedin
+    tags:
+      - always
+
+  - name: Get hosts
+    ovirt_hosts_facts:
+      auth: "{{ ovirt_auth }}"
+      pattern: "cluster={{ cluster_name | mandatory }} and update_available=true"
```
mwperina: I think we should limit hosts:

1. By status
    - we should provide role parameter to override host statuses in which hosts should be upgraded
    - by default I think we should try to upgrade only hosts in Up, because other status usually require administrators attention
2. By host name
    - by default this parameter will be empty (meaning that all hosts can be upgraded), or if non-empty then only hosts mentioned in this list will be upgraded

roles/ovirt-cluster-upgrade/tasks/upgrade.yml
```diff
@@ -0,0 +1,20 @@
+---
+- block:
+
+  - name: Upgrade host
+    ovirt_hosts:
+      auth: "{{ ovirt_auth }}"
+      name: "{{ item.name }}"
+      state: upgraded
+      timeout: 1200
```
mwperina: This should be a role parameter, we cannot assume any value here

mwperina (changes requested):

roles/ovirt-cluster-upgrade/README.md
mykaul: OpenShift is not a product name - the project uses the same name. It is also much more meaningful than my-vm-*.
What does openshift-ansible use? Let's be consistent with whatever it uses by default.

roles/ovirt-cluster-upgrade/tasks/main.yml
```diff
@@ -0,0 +1,128 @@
...
+    - name: Get name of the original scheduling policy
+      ovirt_scheduling_policies_facts:
+        auth: "{{ ovirt_auth }}"
+        id: "{{ ovirt_clusters[0].scheduling_policy.id }}"
+      tags:
+        - ovirt-cluster-upgrade
+  
+    - name: Remember the cluster scheduling policy
+      set_fact:
+        cluster_sched_policy: "{{ ovirt_scheduling_policies[0].name }}"
```
masayag: Could we use a complete word, i.e. `cluster_scheduling_policy`

roles/ovirt-cluster-upgrade/tasks/main.yml
```diff
@@ -0,0 +1,128 @@
...
+      tags:
+        - ovirt-cluster-upgrade
+  
+    - name: Get list of VMs in cluster
+      ovirt_vms_facts:
+        auth: "{{ ovirt_auth }}"
+        pattern: "cluster=production"
```
masayag: Shouldn't the cluster name be passed via variable  "{{ cluster_name }}" ?

roles/ovirt-cluster-upgrade/README.md
```diff
@@ -0,0 +1,57 @@
...
+ * Ansible version 2.3 or higher
+ * Python SDK version 4 or higher
+
+Role Variables
+--------------
+
+| Name               | Default value         |                            |
+|--------------------|-----------------------|----------------------------| 
+| cluster            | Default               | Name of the cluster to be upgraded |
+| low_prio_vms       | UNDEF                 | List of VM names to be stopped before upgrade |
```
masayag: How about allowing to provide a pattern for the vms to be stopped instead of a static list ?

The name `low_prio_vms` isn't clear enough without reading the explanation of it. I'd suggest simply `stopped_vms`

roles/ovirt-cluster-upgrade/README.md
```diff
@@ -0,0 +1,57 @@
...
+| force_pinned       | false                 | If true the pinned to host VMs will be stopped |
```
masayag: why this should be a specific case ? I'd consider HA vms to be more important in that sense.

I'd generalize it and add `force` flag to stop all of the VMs and upgrade the hosts.
HA vms should be started once the upgrade is completed.

roles/ovirt-cluster-upgrade/tasks/main.yml
```diff
@@ -0,0 +1,128 @@
...
+      msg: "No hosts to be updated"
+    when: ovirt_hosts | length == 0
+  
+  - block:
+    - name: Get cluster facts
+      ovirt_clusters_facts:
+        auth: "{{ ovirt_auth }}"
+        pattern: "name={{ cluster_name }}"
+      tags:
+        - ovirt-cluster-upgrade
```
masayag: why it is needed to tag all of the tasks with 'ovirt-cluster-upgrade' if they are already part of the 'ovirt-cluster-upgrade' role ?

masayag (changes requested):

roles/ovirt-cluster-upgrade/tasks/main.yml
masayag: There is an additional issue with this one: The 'upgraded' state expects the host to be in 'UP' status after the upgrade is completed. However, according to [UpgradeHostInternalCommand](https://github.com/oVirt/ovirt-engine/blob/master/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/hostdeploy/UpgradeHostInternalCommand.java#L135), if the upgrade was executed for host in maintenance mode, it is expected to return to that state after the upgrade is done.

1. Where are the timeouts for host status polling or for jobs polling is set ?
2. What is the concurrency of clusters upgrade ? According to this one, we set no boundaries for that, so we should be limited by ansible default sessions. I think we should add some control for that on the engine side.

`state == 'upgraded':
            ret = hosts_module.action(
                action='upgrade',
                action_condition=lambda h: h.update_available,
                wait_condition=lambda h: h.status == hoststate.UP,
                fail_condition=failed_state,
            )`

roles/ovirt-cluster-upgrade/README.md
machacekondra: You may not cover all the VMs you want to shutdown by one pattern. I agree with `stopped_vms`.

roles/ovirt-cluster-upgrade/README.md
machacekondra: Most of their roles are not documented. Also most of their roles are different, it's mostly provisioning of hosts, those are mostly buidling of infra. But if I take an example of firewalld they use name like httpd, and specific number, instead of `myservice` and port `1234`.

roles/ovirt-cluster-upgrade/library/ovirt_scheduling_policies_facts.py
machacekondra: Done: https://github.com/ansible/ansible/pull/25055

roles/ovirt-cluster-upgrade/tasks/main.yml
machacekondra: I've read it's good practice. But I don't see the reason for it as well, it's OK to tag the role in the playbook, if we want to run just one role in a playbook, where there is more roles. I will remove it.

roles/ovirt-cluster-upgrade/tasks/main.yml
machacekondra: Yes, thanks.

roles/ovirt-cluster-upgrade/tasks/upgrade.yml
machacekondra: Done

roles/ovirt-cluster-upgrade/tasks/pinned_vms.yml
machacekondra: It is optional, if user set `force_pinned=false` (the default) we simply skip it.

roles/ovirt-cluster-upgrade/tasks/main.yml
machacekondra: done

roles/ovirt-cluster-upgrade/tasks/main.yml
machacekondra: done

roles/ovirt-cluster-upgrade/README.md
machacekondra: Because those VMs can't be migrated, so user should say, if he want to skip the upgrade of such a host or stop the VMs.

roles/ovirt-cluster-upgrade/tasks/main.yml
machacekondra: 1. There is `upgrade_timeout` variable, which default value is 1200.
2. Not sure I understand. There is no concurency. If using this role, you should use just one target host (usually engine or localhost) to execute it. Concurency would be in this case quite complex, because of migration, so I think we should for now keep it sequantial.

Right, the condition have to be fixed. Thank you.

machacekondra:
Verified, see:

 https://asciinema.org/a/122760

roles/ovirt-cluster-upgrade/README.md
```diff
@@ -0,0 +1,60 @@
...
+ * Python SDK version 4 or higher
+
+Role Variables
+--------------
+
+| Name               | Default value         |                                                     |
+|--------------------|-----------------------|-----------------------------------------------------|
+| cluster_name       | Default               | Name of the cluster to be upgraded.                 |
+| stopped_vms        | UNDEF                 | List of VM names to be stopped before upgrade.      |
+| force_pinned       | false                 | If true the pinned to host VMs will be stopped.     |
```
mwperina: It's no obvious from my side what this parameter does from the 1st look, I'd prefer 'stop_pinned_to_host_vms'. I know it's long :-) but it's not an issue from yml file and everybody know what this parameter does from its name.

roles/ovirt-cluster-upgrade/library/ovirt_clusters.py
```diff
@@ -0,0 +1,567 @@
+#!/usr/bin/python
```
mwperina, line 1: Let's make a note somewhere, that we should remove this module from the role once the module update is backported to 2.3.1 and bump dependency to Ansible 2.3.1

roles/ovirt-cluster-upgrade/library/ovirt_hosts.py
```diff
@@ -0,0 +1,464 @@
+#!/usr/bin/python
```
mwperina, line 1: Let's make a note somewhere, that we should remove this module from the role once the module update is backported to 2.3.1 and bump dependency to Ansible 2.3.1

roles/ovirt-cluster-upgrade/library/ovirt_scheduling_policies_facts.py
mwperina: Let's make a note somewhere, that we should remove this module from the role once we bump dependency to Ansible 2.4.0

mwperina (changes requested):

roles/ovirt-cluster-upgrade/library/ovirt_clusters.py
machacekondra: I will add library/README.md file which will explain those modules.

roles/ovirt-cluster-upgrade/library/ovirt_hosts.py
machacekondra: I will add library/README.md file which will explain those modules.

roles/ovirt-cluster-upgrade/README.md
machacekondra: Done

mwperina (approved):
👍

PR #4, opened 2017-05-18
Add role for ManageIQ/CloudForms deployment

machacekondra:
Add role for ManageIQ/CloudForms deployment

### DIFF
--- a/roles/ovirt-manageiq/README.md
+++ b/roles/ovirt-manageiq/README.md
@@ -1,0 +1,111 @@
+ManageIQ/CloudForms deployment on top of oVirt/RHV
+==================================================
+
+This role downloads ManageIQ/CloudForms qcow image and deploys it into oVirt/RHV.
+Create a virtual machine and attach the ManageIQ/CloudForms disk. Then wait for
+initialization of the ManageIQ/CloudForms system and register the RHV/oVirt as
+a infrastructure provider.
+
+Requirements
+------------
+
+oVirt has to be 4.0.4 or higher and [ovirt-imageio](http://www.ovirt.org/develop/release-management/features/storage/image-upload/) must be installed and running.
+The target machine has to have installed [Python SDK](https://pypi.python.org/pypi/ovirt-engine-sdk-python/4.0.4).
+
+Check `ovirt-imageio-proxy` is running on engine:
+
+```
+systemct status ovirt-imageio-proxy
+```
+
+Also check if `ovirt-imageio-daemon` is running on hosts:
+
+```
+systemct status ovirt-imageio-daemon
+```
+
+You also have to use your CA certificate of the engine.
+You must configure the `ovirt_ca` varible with the path to the CA.
+
+Role Variables
+--------------
+
+QCOW varibles:
+| Name          | Default value                                            |  Description                                                 |
+|---------------|----------------------------------------------------------|--------------------------------------------------------------|
+| miq_qcow_url  | http:////releases.manageiq.org/manageiq-ovirt-fine-1.qc2 | Define the URL where the ManageIQ/CloudForms qcow is stored. |
+
+oVirt/RHV engine login variables:
+| Name                | Default value     |  Description                                    |
+|---------------------|-------------------|-------------------------------------------------|
+| engine_user         | UNDEF             | Define the user to be used to access oVirt/RHV. |
+| engine_password     | UNDEF             | Define password of the 'ovirt_user'.            |
+| engine_fqdn         | UNDEF             | Define the FQDN of the oVirt/RHV.               |
+| engine_ca           | UNDEF             | Define the path to the CA of the oVirt/RHV.     |
+
+Virtual Machine variables:
+| Name               | Default value     |  Description                                 |
+|--------------------|-------------------|----------------------------------------------|
+| miq_vm_name        | manageiq_fine     | Define the name of the ManageIQ/CloudForms VM in oVirt/RHV. |
+| miq_vm_cluster     | Default           | Define the cluster of the virtual machine.    |
+| miq_vm_memory      | 6GiB              | Define the system memory of the virtual machine.    |
+| miq_vm_cpu         | 2                 | Define the number of virtual machine CPU cores.   |
+| miq_vm_os          | rhel_7x64         | Define the operating system of the virtual machine. |
+| miq_vm_cloud_init  | UNDEF             | Define cloud init dictionary to be passed to virtual machine. |
+
+Virtual Machine disks variables:
+| Name                | Default value     |  Description                                         |
+|---------------------|-------------------|------------------------------------------------------|
+| miq_vm_disk_name    | `miq_vm_name`     | Define name of the virtual machine disk.             | 
+| miq_vm_disk_storage | UNDEF             | Define the target storage domain of the disk.        |
+| miq_vm_disk_size    | 50GiB             | Define the size of the virtual machine disk.         |
+| miq_vm_disk_interface | virtio          | Define the interface of the virtual machine disk.    |
+| miq_vm_disk_format  | cow               | Define the format of the virtual machine disk.       |
+
+Virtual Machine nics variables:
+| Name                | Default value     |  Description                                         |
+|---------------------|-------------------|------------------------------------------------------|
+| miq_vm_nics         | {'name': 'nic1', 'profile_name': 'ovirtmgmt', 'interaface': 'virtio'} | Dictionary variable which define network interfaces for virtual machine. |
+
+RHV provider variables:
+| Name                  | Default value     |  Description                                         |
+|-----------------------|-------------------|------------------------------------------------------|
+| miq_rhv_provider_name | RHV provider      | Name of the RHV provider to be displayed in ManageIQ.|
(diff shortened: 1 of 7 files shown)


machacekondra:
@mwperina

machacekondra:
@mgoldboi @oourfali

roles/ovirt-manageiq/README.md
```diff
@@ -0,0 +1,105 @@
...
+|--------------------|-------------------|-------------------------------------------------|
+| ovirt_user         | UNDEF             | Define the user to be used to access oVirt/RHV. |
+| ovirt_password     | UNDEF             | Define password of the 'ovirt_user'.            |
+| ovirt_fqdn         | UNDEF             | Define the FQDN of the oVirt/RHV.               |
+| ovirt_ca           | UNDEF             | Define the path to the CA of the oVirt/RHV.     |
+
+Virtual Machine variables:
+| Name               | Default value     |  Description                                 |
+|--------------------|-------------------|----------------------------------------------|
+|  miq_vm_name       | manageiq_euwe     | Define the name of the ManageIQ/CloudForms VM in oVirt/RHV.|
```
oourfali: remove the extra space.

roles/ovirt-manageiq/tasks/main.yml
```diff
@@ -0,0 +1,126 @@
...
+  register: downloaded
+
+- name: Check file type
+  command: "/usr/bin/file {{ miq_image_path | quote }}"
+  changed_when: false
+  register: filetype
+
+- name: Fail if image is not qcow
+  fail:
+    msg: "The downloaded file is not valid QCOW file."
```
oourfali: "a valid"

roles/ovirt-manageiq/tasks/main.yml
```diff
@@ -0,0 +1,126 @@
...
+      type: server
+      disks:
+        - id: "{{ ovirt_disk.id }}"
+          bootable: true
+      cloud_init: "{{ miq_vm_cloud_init | default(omit) }}"
+      nics: "{{ miq_vm_nics }}"
+    when: ovirt_vms | length == 0
+    register: create_vm
+
+  - name: Wait for VM IP
```
oourfali, line 77: does the image include an ovirt guest agent?

oourfali (commented):
Also I wonder whether we should also add DWH properties (so that capacity and utilization will be configured as well).
Do we have these details? Shall we ask the user fo fill them in?

roles/ovirt-manageiq/README.md
```diff
@@ -0,0 +1,105 @@
+oVirt/RHV ManageIQ/CloudForms deployment
```
mwperina: I'd rephrase as "ManageIQ/CloudForms deployment on top of oVirt/RHV"

roles/ovirt-manageiq/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+You also have to use your CA certificate of the engine.
+You must configure the `ovirt_ca` varible with the path to the CA.
+
+Role Variables
+--------------
+
+QCOW varibles:
+```
+ - miq_qcow_url - Define the URL where the ManageIQ/CloudForms qcow is stored (default: http://releases.manageiq.org/manageiq-ovirt-euwe-1.ova).
```
mwperina: Here we have a link to OVA, below you are using link to qcow2, so shouldn't use single term everywhere and just mention in the doc that both OVA or qcow2 formats can be used?

roles/ovirt-manageiq/README.md
```diff
@@ -0,0 +1,105 @@
...
+| ovirt_ca           | UNDEF             | Define the path to the CA of the oVirt/RHV.     |
+
+Virtual Machine variables:
+| Name               | Default value     |  Description                                 |
+|--------------------|-------------------|----------------------------------------------|
+|  miq_vm_name       | manageiq_euwe     | Define the name of the ManageIQ/CloudForms VM in oVirt/RHV.|
+| miq_vm_cluster     | Default           | Define the cluster of the virtual machine    |
+| miq_vm_memory      | 6GiB              | Define the system memory of the virtual machine    |
+| miq_vm_cpu         | 2                 | Define the number of virtual machine CPU cores   |
+| miq_vm_os          | rhel_7x64         | Define the operating system of the virtual machine |
```
mwperina: The default is upstream, should we should use CentOS instead of RHEL

roles/ovirt-manageiq/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+Note that for passwords you should be using Ansible vault.
+
+```yaml
+    - name: Deploy ManageIQ to oVirt engine
+      hosts: localhost
+      gather_facts: no
+
+      vars:
+        ovirt_fqdn: ondra.local
```
mwperina: ovirt-engine.example.com

roles/ovirt-manageiq/README.md
```diff
@@ -0,0 +1,105 @@
...
+        ovirt_password: 123456
+        ovirt_username: admin@internal
+
+        miq_vm_name: manageiq_euwe
+        miq_qcow_url: http://releases.manageiq.org/manageiq-ovirt-euwe-3.qc2
```
mwperina: Does ManageIQ provide some link to latest release, so we won't need to update this often?

roles/ovirt-manageiq/README.md
```diff
@@ -0,0 +1,105 @@
...
+        miq_vm_cluster: mycluster
+        miq_vm_cloud_init:
+          user_name: root
+          root_password: securepassword
+
+      roles:
+        - machacekondra.ovirt-manageiq
```
mwperina: Shouldn't there be only "ovirt.manageiq" or "ovirt-ansible.ovirt-manageiq"?

roles/ovirt-manageiq/defaults/main.yml
```diff
@@ -0,0 +1,30 @@
...
+miq_password: smartvm
+miq_provider_name: RHV provider
+
+### oVirt/RHV ###
+# VM variables:
+miq_vm_name: manageiq_euwe
+miq_vm_cluster: Default
+miq_vm_memory: 6GiB
+miq_vm_cpu: 2
+miq_vm_os: rhel_7x64
```
mwperina: CentOS?

mwperina (changes requested):

mgoldboi:
would be great if we can add the option to configure additional providers as well as part of the deployment/ configuration stage.
for the use case this CF engine is managing a different provider than oVirt. Openshift for example can be nice part of this story.

mwperina:
> would be great if we can add the option to configure additional providers as well as part of the 
> deployment/ configuration stage.
> for the use case this CF engine is managing a different provider than oVirt. Openshift for example can 
> be nice part of this story.

Not sure if this part of this role, isn't a role configuring OpenShift provider in MiQ part of OpenShift roles, so users could reuse it?

roles/ovirt-manageiq/tasks/main.yml
machacekondra: The ManageIQ includes them, I didn't try the CloudForms.

roles/ovirt-manageiq/README.md
machacekondra: There is no such option.

roles/ovirt-manageiq/README.md
machacekondra: I didn't find any.

roles/ovirt-manageiq/README.md
machacekondra: Yes.

roles/ovirt-manageiq/README.md
machacekondra: Added the qcow only.

roles/ovirt-manageiq/README.md
mwperina: Ahh, so this is oVirt specific, we need to set VM OS to RHEL, even though CentOS is used for ManageIQ VM.

mwperina (approved):
Looks good to me, thanks!

machacekondra:
Added Openshift and DWH support.

mykaul:
Openshift is unrelated to Manage IQ?

machacekondra:
It's related, you can for example manage Openshift in RHV all via ManageIQ. This was requested by Moran.

mykaul:
Makes little sense to me.
I see 2-3 roles:
- Install ManageIQ
- Install OpenShift
- Connect the two.

It's not natural to see it in a single role.

oourfali:
I also remember we agreed to keep openshift out of the scope of this role.
Am I wrong?

machacekondra:
It's optional. It's just one API call, so no problem to remove it.

Yes, that could be separated to those roles, I think better it would be if this role just would deploy ManageIQ, register RHV and then the user can use some ManageIQ modules( if any ) and register other providers in post_tasks.

I will split it.

mwperina (changes requested):
Let's remove Openshift part from this role, it's should be of a different role. Other than that I think it's ready to merge

mwperina (approved):
👍

PR #5, opened 2017-06-14
Update requirements to include python JMESPath

cwilkers:
Reason:
https://github.com/oVirt/ovirt-ansible/blob/master/roles/ovirt-image-template/tasks/main.yml#L46

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 ansible
 yamllint
-python-jmespath
+jmespath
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -1,2 +1,3 @@
 ansible
 yamllint
+python-jmespath
```
machacekondra: It's actually `jmespath`, without `python` prefix.

machacekondra:
Thank you for PR. Can you please check my comment. Thank you.
CC @mwperina

cwilkers:
Thank you @machacekondra , I misunderstood the role of requirements.txt.

mwperina (approved):

PR #16, opened 2017-07-19
Edited the ovirt-infra role.

bgraveno:
Hi @machacekondra,

Can you please review and merge this request?

Question: We use the term "variables" in this content. However when viewing the dependent role, we use the term "item"?  For example in oVirt Networks we say, "The item in logical_networks list can contain following parameters.." Should I change that to variable as well?

Kind regards,

### DIFF
--- a/roles/ovirt-infra/README.md
+++ b/roles/ovirt-infra/README.md
@@ -1,7 +1,7 @@
-oVirt infra
+oVirt Infra
 =========
 
-This role setup oVirt infrastructure including: data centers, clusters, networks, hosts, users and groups.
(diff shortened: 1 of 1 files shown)


mwperina (approved):
👍

machacekondra:
> Question: We use the term "variables" in this content. However when viewing the dependent role, we use the term "item"? For example in oVirt Networks we say, "The item in logical_networks list can contain following parameters.." Should I change that to variable as well?

Well, the variable is `logical_networks`, which is a list, and every item in that list is usually called `item`, so I hope it's clear, but if it's not please rephrase it.

bgraveno:
Thanks, that clears it up.

PR #17, opened 2017-07-19
ovirt-storages: expose override_luns parameter

tareqalayan:
This param allow us to format iscsi luns when we add iscsi storage to
engine.

### DIFF
--- a/roles/ovirt-storages/tasks/main.yml
+++ b/roles/ovirt-storages/tasks/main.yml
@@ -16,6 +16,7 @@
     posixfs: "{{ item.value.posixfs | default(omit) }}"
     glusterfs: "{{ item.value.glusterfs | default(omit) }}"
     fcp: "{{ item.value.fcp | default(omit) }}"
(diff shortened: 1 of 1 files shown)


roles/ovirt-storages/tasks/main.yml
```diff
@@ -16,6 +16,7 @@
     posixfs: "{{ item.value.posixfs | default(omit) }}"
     glusterfs: "{{ item.value.glusterfs | default(omit) }}"
     fcp: "{{ item.value.fcp | default(omit) }}"
+    override_luns: "{{ item.value.override_luns | default(omit) }}"
```
machacekondra, line 19: I think it's part only of iscsi, no? Also please document it in `README.md`. Thanks!

machacekondra: Actually if it's part of `iscsi` param there is no reason for PR.

tareqalayan:
i failed on adding the iscsi in the first time .. will close it now ...thanks and sorry for the noise

PR #33, opened 2017-07-25
Add prefix for roles in RPM

machacekondra:
This PR add a prefix for roles in RPM, according to platform guidlines.


machacekondra:
After consultation, we keep the naming without prefix. Closing.

mwperina:
All of our roles contains 'ovirt-' prefix in their names, so they are clearly identified in target directory (for example /usr/share/ansible/roles/ovirt-infra). RHEL system role don't (for example kdump or selinux), so that's why they want to add unifying prefix into directory name (for example /usr/share/ansible/roles/rhel-system-role.kdump). 

It seems to me to have too much 'ovirt' if we use the same (for example /usr/share/ansible/roles/ovirt.ovirt-infra)

PR #51, opened 2017-08-23
pass connection type to storage_connections module

tareqalayan:
I encountered this issue again : oVirt#38
This should fix it.


machacekondra:
`should`? Can you reproduce the issue without the patch? And don't reproduce with the patch?

machacekondra:
add to whitelist

machacekondra:
This PR doesn't solve the real issue, closing.

PR #52, opened 2017-08-23
ovirt-clusters: set ksm & balooning to be True

### LINKED ISSUE #45: unsuitable profile default in ovirt-cluster
Production should not disable ksm nor ballooning

tareqalayan:
fixes: https://github.com/oVirt/ovirt-ansible/issues/45

### DIFF
--- a/roles/ovirt-clusters/README.md
+++ b/roles/ovirt-clusters/README.md
@@ -59,8 +59,8 @@ Possible `profile` options are `development` and `production`, their default val
 
 | Parameter                         | Value              |
 |-----------------------------------|--------------------|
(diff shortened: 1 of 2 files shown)


mykaul:
Is this the default when creating a new cluster? If not (and I suspect it isn't), why are you changing it here?

tareqalayan:
well, i wanted to close this issue https://github.com/oVirt/ovirt-ansible/issues/45  
By default KSM and ballooning are not enabled. 
But the profile name called Production, it may implies non default settings.
If you think that this shouldn't be changed i will close this pull.

mykaul:
Closed #45 .

PR #53, opened 2017-08-24
Remove firewalld supported parameter and no longer needed task cluster_version_4.2.yml

mwperina:


mwperina:
@machacekondra 

Verified with local installation for adding new host, reinstalling host and re-adding previously removed host

machacekondra (approved):

PR #56, opened 2017-08-28
Shorten VDSM RPM version fetch

mykaul:
You could use RPM query format to get the version, no need for grep and awk.


mwperina (approved):
👍

machacekondra (approved):
Thanks

PR #64, opened 2017-08-30
ovirt-cluster-upgrade: Don't run check_for upgrade if not needed

machacekondra:
This PR add new `check_upgrade` parameter.

If this parameter is `true`, we will fetch all host according to user's pattern, even host which don't report available updates and then check for upgrade action, before performing actual upgrade action.

If this parameter is `false`, we only fetch hosts which has update_available, and we don't check for upgrade.

### DIFF
--- a/roles/ovirt-cluster-upgrade/README.md
+++ b/roles/ovirt-cluster-upgrade/README.md
@@ -20,6 +20,7 @@ Role Variables
 | upgrade_timeout         | 1200                  | Timeout in seconds to wait for host to be upgraded. |
 | host_statuses           | [UP]                  | List of host statuses. If a host is in any of the specified statuses then it will be upgraded. |
 | host_names              | [\*]                  | List of host names to be upgraded.        |
+| check_upgrade           | false                 | If true, run check_for_upgrade action on all hosts before executing upgrade on them. If false, run upgrade only for hosts with available upgrades and ignore all other hosts. |
 
 Dependencies
 ------------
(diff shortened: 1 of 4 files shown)


machacekondra:
@mwperina

roles/ovirt-cluster-upgrade/README.md
```diff
@@ -20,6 +20,7 @@ Role Variables
 | upgrade_timeout         | 1200                  | Timeout in seconds to wait for host to be upgraded. |
 | host_statuses           | [UP]                  | List of host statuses. If a host is in any of the specified statuses then it will be upgraded. |
 | host_names              | [\*]                  | List of host names to be upgraded.        |
+| check_upgrade           | false                 | If true, run check_for_upgrade action on all hosts before trying to upgrade them. If false, we only upgrade hosts which reporsts available updates. |
```
mwperina: If `true`, run check_for_upgrade action on all hosts before executing upgrade on them. If `false`, run upgrade only for hosts with available upgrades and ignore all other hosts.

mwperina (changes requested):
Looks good, just minor doc change

machacekondra:
ci build please

roles/ovirt-cluster-upgrade/README.md
machacekondra: done

mwperina:
👍

PR #69, opened 2017-09-03
Support DR with oVirt

maorlipchuk:
Support DR with oVirt

This patch should add support for Disaster Recovery with oVirt
using an ansible tool which can recover and clean the oVirt setup.

The current script supports ansible latest master (v2.5 future
release)
The SDK version this script is based on is
   python-ovirt-engine-sdk4-4.2.1-1.a2.20171031git7f3b968.fc25.x86_64.rpm

To execute the DR, one can use the following command:
~/git/ansible/bin/ansible-playbook ~/ovirt-ansible/examples/dr_ovirt_setup.yml  --tags "fail_back" --ask-vault-pass -vvv

   {ansible_home}/bin/ansible-playbook ./ovirt-ansible/examples/dr_ovirt_setup.yml
        --extra-vars "target_host='primary'" --ask-vault-pass

Here is an example of the var file that should be used:

There are extra vars which can be configured as part of the dr support:
  target_host    - Indicates which host should the ansible script be executed (Optional values are primary and secondary).
  source_mapping - Indeicate the mapping which will be used to map each value of the entity registered in the target host.

Here are the following tags which are supported:
 - fail_back - Supports fail back scenrio from site at source map to the destination site mentioned in target_host
 - fail_over- Supports fail over scneario from site mentioned in target_host to site mentioned in source_map
 - clean_engine - Cleans the engine setup of the source_map

### DIFF
--- a/examples/disaster_recovery_vars.yml
+++ b/examples/disaster_recovery_vars.yml
@@ -1,0 +1,96 @@
+---
+dr_sites_primary_url: "http://mlipchuk.usersys.redhat.com:8080/ovirt-engine/api"
+dr_sites_primary_username: "admin@internal"
+dr_sites_primary_ca_file: "/home/mlipchuk/ovirt-share/etc/pki/ovirt-engine/ca.pem"
+
+dr_sites_secondary_url: "http://10.35.0.140:8080/ovirt-engine/api"
+dr_sites_secondary_username: "admin@internal"
+dr_sites_secondary_ca_file: "/home/mlipchuk/ovirt-share/etc/pki/ovirt-engine/ca.pem"
+
+dr_import_storages:
+    - dr_domain_type: "nfs"
+      dr_master_domain: "True"
+      dr_primary_address: "10.35.16.43"
+      dr_primary_path: "/export/data_test7"
+      dr_primary_dc_name: "Prod"
+      dr_primary_name: "data7"
+      dr_secondary_address: "10.35.1.68"
+      dr_secondary_path: "/home/export/data2"
+      dr_secondary_dc_name: "Recovery"
+      dr_secondary_name: "data2"
+
+    - dr_domain_type: "nfs"
+      dr_master_domain: "False"
+      dr_primary_name: "data6"
+      dr_primary_address: "10.35.16.43"
+      dr_primary_path: "/export/data_test6"
+      dr_primary_dc_name: "Prod"
+      dr_secondary_name: "data"
+      dr_secondary_address: "10.35.1.68"
+      dr_secondary_path: "/home/export/data1"
+      dr_secondary_dc_name: "Recovery"
+
+    - dr_domain_type: "iscsi"
+      dr_master_domain: "False"
+      dr_primary_id: "aa92cc71-1b88-4998-a755-970ef8a638ea"
+      dr_primary_address: "10.35.16.55"
+      dr_primary_target: ["iqn.2015-07.com.mlipchuk1.redhat:444"]
+      dr_primary_dc_name: "Prod"
+      dr_primary_name: "scsi2"
+      dr_secondary_id: "ccccf47b-ff77-4660-bf07-2809a750adb6"
+      dr_secondary_address: "10.35.0.222"
+      dr_secondary_target: ["iqn.2015-07.com.recovery.redhat:444"]
+      dr_secondary_dc_name: "Recovery"
+      dr_secondary_name: "scsi2"
+
+# Mapping for cluster
+dr_cluster_mappings:
+    - primary_name: cluster_prod
+      secondary_name: cluster_recovery
+
+# Mapping for affinity group
+dr_affinity_group_mappings:
+    - primary_name: "primary_affinity"
+      secondary_name: "secondary_affinity"
+
+# Mapping for affinity label
+dr_affinity_label_mappings:
+    - primary_name: "label_prod"
+      secondary_name: "label_recovery"
+
+# Mapping for domain
+dr_domain_mappings:
+    - primary_name: "new-authz"
+      secondary_name: "internal-authz"
+
+# Mapping for roles
+dr_role_mappings:
+    - primary_name: "VmMananger"
+      secondary_name: "NeverMnd"
+
+# Mapping for vnic profile
+dr_network_mappings:
+    - primary_network_name: "ovirtmgmt"
+      primary_profile_name: "ovirtmgmt"
+      primary_profile_id: "e368cbd4-59d9-4a7e-86c1-e405c916a836"
+      secondary_network_name: "ovirtmgmt"
+      secondary_profile_name: "ovirtmgmt"
+      secondary_profile_id: "e368cbd4-59d9-4a7e-86c1-e405c916a836"
+
+# Mapping for direct LUN disks
+dr_lun_mappings:
+    - primary_logical_unit_id: "360014056a2be431c0fd46c4bdce92b66"
+      primary_logical_unit_address: "10.35.16.55"
+      primary_logical_unit_port: 3260
+      primary_logical_unit_portal: "1"
+      primary_logical_unit_username: ""
+      primary_logical_unit_password: ""
+      primary_logical_unit_target: "iqn.2015-07.com.mlipchuk1.redhat:444"
+      secondary_storage_type: "iscsi"
+      secondary_logical_unit_id: "36001405961a7f95e6aa461b8dba53052"
+      secondary_logical_unit_address: "10.35.0.222"
+      secondary_logical_unit_port: 3260
+      secondary_logical_unit_portal: "1"
+      secondary_logical_unit_username: ""
+      secondary_logical_unit_password: ""
+      secondary_logical_unit_target: "iqn.2015-07.com.recovery.redhat:444"
--- a/examples/dr_ovirt_setup.yml
+++ b/examples/dr_ovirt_setup.yml
@@ -1,0 +1,9 @@
+---
+- name: Setup oVirt environment
+  hosts: localhost
+  connection: local
+  vars_files:
+     - ovirt_passwords.yml
+     - disaster_recovery_vars.yml
+  roles:
+     - {role: ovirt-disaster-recovery}
--- a/roles/ovirt-disaster-recovery/defaults/main.yml
+++ b/roles/ovirt-disaster-recovery/defaults/main.yml
@@ -1,0 +1,14 @@
+# Indicate whether to ignore errors on clean engine setup.
+dr_ignore_error_clean: "no"
+
+# Indicate whether to ignore errors on recover.
+dr_ignore_error_recover: "yes"
+
+# Indicate whether to use the partial import flag when registering VMs and Templates.
+dr_partial_import: "True"
+
+# Indicate the default target host to be used in the play
+dr_target_host: "primary"
+
+# Indicate the default sourece map o be used in the play
+dr_source_map: "secondary"
--- a/roles/ovirt-disaster-recovery/tasks/clean/remove_disks.yml
+++ b/roles/ovirt-disaster-recovery/tasks/clean/remove_disks.yml
@@ -1,0 +1,5 @@
+- name: Remove disk
+  ovirt_disks:
+      state: absent
+      id: "{{ disk.id }}"
+      auth: "{{ ovirt_auth }}"
--- a/roles/ovirt-disaster-recovery/tasks/clean/remove_domain.yml
+++ b/roles/ovirt-disaster-recovery/tasks/clean/remove_domain.yml
@@ -1,0 +1,11 @@
+# If we get the exception "Cannot deactivate Master Data Domain while there are running tasks on its Data Center."
+# We should wait for some time and try again
+- name: Remove storage domain
+  ovirt_storage_domains:
+      state: absent
+      id: "{{ storage.id }}"
+      name: "{{ storage.name }}"
+      auth: "{{ ovirt_auth }}"
+      host: "{{ host }}"
+      destroy: "{{ dr_force }}"
+      data_center: "{{ sp_uuid }}"
--- a/roles/ovirt-disaster-recovery/tasks/clean/remove_domain_process.yml
+++ b/roles/ovirt-disaster-recovery/tasks/clean/remove_domain_process.yml
@@ -1,0 +1,39 @@
+# TODO: Check what happens when we force remove unattached storage domain (probably should add a default empty GUID as a data center
+# Answer: When we force remove an unattached storage domain, ansible tries to move it to maintenance and detach it first, although it might be that this storage domain is already detached and has no related data center, therefor the move to maintenance will fail
+
+# We set an initial value for sp_uuid since this task is being called
+# multiple times from the main task and sp_uuid is stateful.
+- name: Set default boolean value for sp_uuid
+  set_fact: sp_uuid=True
+
+- name: Detached storage domain - Set sp_uuid with empty GUID
+  set_fact: sp_uuid="00000000-0000-0000-0000-000000000000"
+  when: storage.data_centers is not defined
+
+- name: Detached storage domain - Fetch active host for remove
+  ovirt_hosts_facts:
+    pattern: "status=up"
+    auth: "{{ ovirt_auth }}"
+  when: storage.data_centers is not defined
+
+- name: Attached storage domain - Fetch active host for remove
+  ovirt_hosts_facts:
+    pattern: "status=up and storage={{ storage.name }}"
+    auth: "{{ ovirt_auth }}"
+  when: storage.data_centers is defined
+
+# If sp_uuid is still initiated with the default boolean value,
+# that means that there is a data center which the storage domain is attached to it.
+- name: Attached storage domain - Set sp_uuid
+  set_fact: sp_uuid="{{ storage.data_centers[0].id }}"
+  when: sp_uuid
+
+- name: Remove storage domain with no force
+  include_tasks: roles/ovirt-disaster-recovery/tasks/clean/remove_domain.yml host={{ ovirt_hosts[0].id }}
+  when: "ovirt_hosts is defined and ovirt_hosts|length > 0 and not force"
+
+- name: Force remove storage domain
+  include_tasks: roles/ovirt-disaster-recovery/tasks/clean/remove_domain.yml host="00000000-0000-0000-0000-000000000000"
+  #storage={{ storage }} host={{ ovirt_hosts[0] }} sp_uuid=sp_uuid
+  when: "force"
+
--- a/roles/ovirt-disaster-recovery/tasks/clean/remove_vms.yml
+++ b/roles/ovirt-disaster-recovery/tasks/clean/remove_vms.yml
@@ -1,0 +1,5 @@
+- name: Remove diskless VMs
+  ovirt_vms:
+      state: absent
+      name: "{{ vm.name }}"
+      auth: "{{ ovirt_auth }}"
--- a/roles/ovirt-disaster-recovery/tasks/clean/shutdown_vms.yml
+++ b/roles/ovirt-disaster-recovery/tasks/clean/shutdown_vms.yml
@@ -1,0 +1,8 @@
+- name: Shutdown VMs
+  ovirt_vms:
+      state: stopped
+      name: "{{ vms.name }}"
+      force: True
+      wait: False
+      auth: "{{ ovirt_auth }}"
+
--- a/roles/ovirt-disaster-recovery/tasks/clean/update_ovf_store.yml
+++ b/roles/ovirt-disaster-recovery/tasks/clean/update_ovf_store.yml
@@ -1,0 +1,5 @@
+- name: update OVF store for storage domain
+  ovirt_storage_domains:
+      state: update_ovf_store
+      name: "{{ storage.name }}"
+      auth: "{{ ovirt_auth }}"
--- a/roles/ovirt-disaster-recovery/tasks/clean_engine.yml
+++ b/roles/ovirt-disaster-recovery/tasks/clean_engine.yml
@@ -1,0 +1,117 @@
+- block:
+    - name: Obtain SSO token
+      ovirt_auth:
+          url: "{{ vars['dr_sites_' + dr_source_map + '_url'] }}"
+          username: "{{ vars['dr_sites_' + dr_source_map + '_username'] }}"
+          password: "{{ vars['dr_sites_' + dr_source_map + '_password'] }}"
+          ca_file: "{{ vars['dr_sites_' + dr_source_map + '_ca_file'] }}"
+
+    # Get all the running VMs and shut them down
+    - name: Fetch VMs in the setup
+      ovirt_vms_facts:
+          pattern:  status != down
+          auth: "{{ ovirt_auth }}"
+
+    - include_tasks: roles/ovirt-disaster-recovery/tasks/clean/shutdown_vms.yml vms={{ item }}
+      with_items: "{{ ovirt_vms }}"
+
+    - name: Update OVF_STORE disks for active storage domain
+      ovirt_storage_domains_facts:
+          pattern:  status = active and type != glance
+          auth: "{{ ovirt_auth }}"
+
+    - include_tasks: roles/ovirt-disaster-recovery/tasks/clean/update_ovf_store.yml storage={{ item }}
+      with_items:
+          - "{{ ovirt_storage_domains }}"
+
+    # We do not supoport export storage domain
+    # TODO: Export storage domain should be handled by an independant task which will clean the meta data.
+    # We first remove all the non master storage domain.
+    - name: Fetch valid storage domains for detach
+      ovirt_storage_domains_facts:
+          pattern:  status = active or status = maintenance or status = unattached and type != glance
+          auth: "{{ ovirt_auth }}"
+
+    - name: Set force remove flag to false
+      set_fact: dr_force=False
+
+    - include_tasks: roles/ovirt-disaster-recovery/tasks/clean/remove_domain_process.yml storage={{ item }}
+      with_items:
+          - "{{ ovirt_storage_domains }}"
+      when: "not item.master"
+
+    # We use inactive filter only at the end, since we are not sure if there were any storage domains
+    # which became inactive on the process or if there were any on the beginning.
+    - name: Fetch invalid storage domains to remove
+      ovirt_storage_domains_facts:
+          # TODO: Should check also about Cinder
+          pattern: status != active and type != glance
+          auth: "{{ ovirt_auth }}"
+
+    - name: Set force remove flag to true
+      set_fact: dr_force=True
+
+    - include_tasks: roles/ovirt-disaster-recovery/tasks/clean/remove_domain_process.yml storage={{ item }}
+      with_items:
+          - "{{ ovirt_storage_domains }}"
+      when: "not item.master"
+
+    # Try to remove the master storage domains, if we fail we will try later to force remove all
+    # inactive master domains.
+    - name: Fetch valid storage domains for detach master
+      ovirt_storage_domains_facts:
+          pattern:  status = active or status = maintenance
+          auth: "{{ ovirt_auth }}"
+
+    - name: Set force remove flag to false
+      set_fact: dr_force=False
+
+    - include_tasks: roles/ovirt-disaster-recovery/tasks/clean/remove_domain_process.yml storage={{ item }}
+      with_items:
+          - "{{ ovirt_storage_domains }}"
+      when: "item.master"
+
+    # If by any chance the master storage domain is inactive we force remove it.
+    - name: Set force remove flag to true
+      set_fact: dr_force=True
+
+    - name: Fetch invalid storage domains to remove master
+      ovirt_storage_domains_facts:
+          pattern: status != active and type != glance
+          auth: "{{ ovirt_auth }}"
+
+    - include_tasks: roles/ovirt-disaster-recovery/tasks/clean/remove_domain_process.yml storage={{ item }}
+      with_items:
+          - "{{ ovirt_storage_domains }}"
+
+    - name: Fetch leftover storage domains
+      ovirt_storage_domains_facts:
+          pattern: type != glance
+          auth: "{{ ovirt_auth }}"
+
+    # Remove direct LUN disks
+    - name: Fetch leftover direct LUN disks in the setup
+      ovirt_disk_facts:
+          pattern: alias != OVF_STORE
+          auth: "{{ ovirt_auth }}"
+
+    - include_tasks: roles/ovirt-disaster-recovery/tasks/clean/remove_disks.yml disk={{ item }}
+      with_items: "{{ ovirt_disks }}"
+
+    # Remove VMs without disks only if there are no data storage domains left in the setup
+    - name: Fetch leftover VMs in the setup
+      ovirt_vms_facts:
+          auth: "{{ ovirt_auth }}"
+      when: "ovirt_storage_domains | length == 0"
+
+    - include_tasks: roles/ovirt-disaster-recovery/tasks/clean/remove_vms.yml vm={{ item }}
+      with_items: "{{ ovirt_vms }}"
+
+  # Default value is set in role defaults
+  ignore_errors: "{{ dr_ignore_error_clean }}"
+- always:
+    - name: Revoke the SSO token
+      ovirt_auth:
+          state: absent
+          ovirt_auth: "{{ ovirt_auth }}"
+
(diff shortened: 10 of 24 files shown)


machacekondra:
@mwperina

ifireball:
looks like this PR is gonna be a little guinea pig for me, apologies in advance...

ifireball:
ci test please

examples/setup_recover.yml
```diff
@@ -0,0 +1,13 @@
+- name: Setup oVirt environment
+  hosts: localhost
+  connection: local
+  vars:
+     detach: "{{ clean | default(False) }}"
```
machacekondra: Why do we need the `default(False)` here? This is an example, so I think it would be better to just specify:

```yaml
detach: true
recover: true
```

in examples.

roles/clean_engine/tasks/main.yml
```diff
@@ -0,0 +1,84 @@
+- block:
+        - name: Obtain SSO token
```
machacekondra: Can you please 4 spaces for indentation?

roles/clean_engine/tasks/main.yml
```diff
@@ -0,0 +1,84 @@
...
+          ovirt_auth:
+                  url: "{{ url }}"
+                  username: "{{ username }}"
+                  password: "{{ password }}"
+                  ca_file: "{{ ca_file }}"
+
+        # TODO: add a test of ISO and export storage domain.
+        # We first remove all the non master storage domains which are active.
+        - set_fact: master=False
+        - ovirt_storage_domains_facts:
```
machacekondra: Can you please use named tasks everywhere, the playbook has much more readable output then:
```yaml
- name: Fetching active storage domains
  ovirt_storage_domains_facts:
    pattern: status = active
    ....
```

roles/clean_engine/tasks/main.yml
```diff
@@ -0,0 +1,84 @@
...
+                pattern: status = active
+                auth: "{{ ovirt_auth }}"
+
+        - include: ../tasks/maintenance_domain.yml storage={{ item }}
```
machacekondra: The `absent` status of `ovirt_storage_domains` module support to swith the domain to maintanence and detach it before remove, so you don't need this I think. Just use `state: absent`

roles/clean_engine/tasks/remove_domain.yml
```diff
@@ -0,0 +1,16 @@
+# TODO: Check what happens when we force remove unattached storage domain (probably should add a default empty GUID as a data center
+# Answer: When we force remove an unattached storage domain, ansible tries to move it to maintenance and detach it first, although it might be that this storage domain is already detached and has no related data center, therefor the move to maintenance will fail
```
machacekondra: If this is true, this is bug in ansible module and should be fixed, but If I look at the code correctly it shouldn't happen

machacekondra:
add to whitelist

machacekondra (commented):
If I remember correctly you've choose to use the ini file for variables just because it's easier to export it from some CLI program, right?

examples/setup_recover.yml
maorlipchuk: Yes, good point, I will change it.
Actually I was thinking to change the value from boolean to the destination setup that the user will want to clean/recover.
Since there are 2 setups, primary oVirt setup and secondary oVirt setup, when the user will try fail over, in case of a disaster, from the primary to the secondary one will want to clean the primary setup, and when the user will want to fail back again to the primary setup, one will want to clean the secondary setup and recover back the primary setup.

so for example when we will want to call the playbook it thought to dosomething like this:
   ansible-playbook setup_recover.yml -i recovery.ini --extra-vars "recover=Primary clean=Secondary" --ask-vault-pass

roles/clean_engine/tasks/main.yml
maorlipchuk: sure, done

roles/clean_engine/tasks/main.yml
maorlipchuk: sure, done

roles/clean_engine/tasks/main.yml
maorlipchuk: Yep, you got a point there, I will change this, thanks

roles/clean_engine/tasks/remove_domain.yml
maorlipchuk: when using 'absent', ovirt_storage_domains will call storage_domains_module.remove which will call _maintenance, and as part of the _maintenance method it will try to fetch the data center because it uses _attached_sds_service, and since the storage domain is not attached to anything it will fail
This is the failure that I get for using 'absent' on unattached storage domain:
Traceback (most recent call last):
  File "/tmp/ansible_jf3Kkc/ansible_module_ovirt_storage_domains.py", line 511, in main
    host=module.params['host'],
  File "/tmp/ansible_jf3Kkc/ansible_modlib.zip/ansible/module_utils/ovirt.py", line 648, in remove
    self.pre_remove(entity)
  File "/tmp/ansible_jf3Kkc/ansible_module_ovirt_storage_domains.py", line 391, in pre_remove
    self._maintenance(storage_domain)
  File "/tmp/ansible_jf3Kkc/ansible_module_ovirt_storage_domains.py", line 345, in _maintenance
    attached_sds_service = self._attached_sds_service()
  File "/tmp/ansible_jf3Kkc/ansible_module_ovirt_storage_domains.py", line 337, in _attached_sds_service
    dc = dcs_service.service(self._module.params['data_center']).get()
  File "/usr/lib64/python2.7/site-packages/ovirtsdk4/services.py", line 4031, in get
    return self._internal_get(headers, query, wait)
  File "/usr/lib64/python2.7/site-packages/ovirtsdk4/service.py", line 202, in _internal_get


I also uploaded a patch that fix that when using force:
https://github.com/ansible/ansible/pull/28962

machacekondra: It broken from:

 https://github.com/ansible/ansible/pull/28659/files

When using `dc = dcs_service.service(self._module.params['data_center']).get()` we ge 404 instead of `None`,
we need to fix it in module.

This[1] doesn't do a proper fix, I would rather re-fix #28659.

[1] https://github.com/ansible/ansible/pull/28962/files

maorlipchuk: Thanks for clearing that out and sorry for the regression but something here look a bit weird, the method name is called _attached_sds_service that basically should mean that this storage domains must have a DC which it attached to.
so I think that this method should not be called at all for detached storage domains, no?

machacekondra: No problem. Right so if `_attached_sds_service` returns None, it the DC isn't found so we can't return attached storage domain service, no?

maorlipchuk: Proposed a fix in the following link: 
   https://github.com/ansible/ansible/pull/29175

mykaul:
Any updates?

maorlipchuk:
We were recently focused on the oVirt gaps before the beta, regarding the mapping of cluster, external LUN disks, affinity groups, affinity labels, and permissions.
Now we can use the python-SDK through ansible to support register of VM/Template with those mapping and add those changes to this PR.
So I assume that now we will mainly focus on this PR and probably will have an updated PR until next week.

roles/ovirt-clean-engine/tasks/shutdown_vms.yml
```diff
@@ -0,0 +1,10 @@
+- name: Run highly available VMs
```
machacekondra: Run? Should it be stop? Also the HA shouldn't be there I think.

roles/ovirt-clean-engine/tasks/shutdown_vms.yml
```diff
@@ -0,0 +1,10 @@
...
+  ovirt_vms:
+      state: stopped
+      name: "{{ vms.name }}"
+      force: True
+      wait: False
+      auth: "{{ ovirt_auth }}"
+      # TODO: Add support for pin to host: Should the host be predefine?
```
machacekondra: For stopping a VM pinning doesn't have any effect AFAIK.

roles/ovirt-clean-engine/tasks/shutdown_vms.yml
```diff
@@ -0,0 +1,10 @@
...
+      #host: "ovirt_hosts[0].name"
```
machacekondra: For stopping the `host` param doesn't have any effect.

roles/ovirt-clean-engine/tasks/main.yml
```diff
@@ -0,0 +1,82 @@
...
+
+    # TODO: add a test of ISO and export storage domain.
+    # We first remove all the non master storage domain.
+    - name: Fetch valid storage domains for detach
+      ovirt_storage_domains_facts:
+          pattern:  status = active or status = maintenance or status = unattached and type != glance
+          auth: "{{ ovirt_auth }}"
+
+    - name: Set force remove flag to false
+      set_fact: force=False
```
machacekondra: please add new line here

roles/ovirt-clean-engine/tasks/main.yml
```diff
@@ -0,0 +1,82 @@
...
+    # Get all the running VMs and shut them down
+    - name: Fetch VMs in the setup
+      ovirt_vms_facts:
+          pattern:  status != down
+          auth: "{{ ovirt_auth }}"
+
+    - include_tasks: tasks/shutdown_vms.yml vms={{ item }}
+      with_items: "{{ ovirt_vms }}"
+
+    # TODO: add a test of ISO and export storage domain.
```
machacekondra: Unfortunatelly we don't have support to search for it in backend, so you need to do it in code, you can take a look here: 

https://github.com/oVirt/ovirt-ansible/blob/master/roles/ovirt-manageiq/tasks/main.yml#L35

roles/ovirt-clean-engine/tasks/main.yml
```diff
@@ -0,0 +1,82 @@
...
+    # We first remove all the non master storage domain.
+    - name: Fetch valid storage domains for detach
+      ovirt_storage_domains_facts:
+          pattern:  status = active or status = maintenance or status = unattached and type != glance
+          auth: "{{ ovirt_auth }}"
+
+    - name: Set force remove flag to false
```
machacekondra: Hmm, why? Why not have it in variables?

roles/ovirt-clean-engine/tasks/main.yml
```diff
@@ -0,0 +1,82 @@
...
+          auth: "{{ ovirt_auth }}"
+
+    - include_tasks: tasks/shutdown_vms.yml vms={{ item }}
+      with_items: "{{ ovirt_vms }}"
+
+    # TODO: add a test of ISO and export storage domain.
+    # We first remove all the non master storage domain.
+    - name: Fetch valid storage domains for detach
+      ovirt_storage_domains_facts:
+          pattern:  status = active or status = maintenance or status = unattached and type != glance
```
machacekondra: Note that here you will remove all storage domains in all datacenters, maybe there should be also condition for specific DC?

roles/ovirt-clean-engine/tasks/main.yml
```diff
@@ -0,0 +1,82 @@
...
+    # We use inactive filter only at the end, since we are not sure if there were any storage domains
+    # which became inactive on the process or if there were any on the beginning.
+    - name: Fetch invalid storage domains to remove
+      ovirt_storage_domains_facts:
+          # TODO: Should check also about Cinder
+          pattern: status != active and type != glance
+          auth: "{{ ovirt_auth }}"
+
+    - name: Set force remove flag to true
+      set_fact: force=True
```
machacekondra: new line please

roles/ovirt-clean-engine/tasks/shutdown_vms.yml
```diff
@@ -0,0 +1,10 @@
+- name: Run highly available VMs
+  ovirt_vms:
+      state: stopped
+      name: "{{ vms.name }}"
+      force: True
+      wait: False
```
machacekondra: Why `wait=false`?

roles/ovirt-recover/tasks/run_vms.yml
```diff
@@ -0,0 +1,8 @@
+- name: Run highly available VMs
+  ovirt_vms:
+      state: running
+      name: "{{ vms.name }}"
+      wait: False
```
machacekondra: Why `false`?

roles/ovirt-recover/tasks/run_vms.yml
```diff
@@ -0,0 +1,8 @@
...
+      auth: "{{ ovirt_auth }}"
+      # TODO: Add support for pin to host: Should the host be predefine?
```
machacekondra: The scheduler chooses it, I would leave it for it, so please remove these comments.

examples/setup_recover.yml
```diff
@@ -0,0 +1,15 @@
...
+     recover: "{{ recovered | default(False) }}"
+     target_host: "{{ host | default('primary') }}"
+     source_map: "{{ source_mapping | default('secondary') }}"
+  vars_files:
+     - group_vars/ovirt_password.yml
```
machacekondra: can you please add those files to this directory as well, so user has an example variables file.

roles/ovirt-clean-engine/defaults/main.yml
```diff
@@ -0,0 +1,11 @@
+# Should perform clean on the setup - Remove all storage domains.
+clean: "False"
+
+# Should the ansible play ignore errors on the way or not.
+ignore: "yes"
```
machacekondra: Can you please prefix all variables with `dr_` prefix? It's considered to be a best prefix. I know some our roles don't follow it, but I would like to if new roles will follow that.

Also I would rename this to `dr_ignore_errors`

roles/ovirt-clean-engine/tasks/main.yml
```diff
@@ -0,0 +1,99 @@
...
+      ovirt_storage_domains_facts:
+          pattern:  status = active or status = maintenance or status = unattached and type != glance
+          auth: "{{ ovirt_auth }}"
+
+    - include_tasks: ../tasks/update_ovf_store.yml storage={{ item }}
+      with_items:
+          - "{{ ovirt_storage_domains }}"
+
+    - name: Set force remove flag to false
+      set_fact: force=False
```
machacekondra: add new line here please

roles/ovirt-recover/defaults/main.yml
```diff
@@ -0,0 +1,14 @@
+# Should perform a recover of the setup using the ini file.
+recovered: "False"
```
machacekondra: Can we please use `dr_` prefix for all variables?

machacekondra (commented):
Can you please not use upper case characters in variable names? like `DR_site`, `import_NFS`, can you please use rather `dr_site`, `import_nfs`, also can you please add `dr_` prefix to all top level variables? like `dr_import_nfs`?

machacekondra:
Many people don't like nested structure for variable, and the Ansible way is using as flat structure as possible, so if there is always one primary and one secondary, you can restructure the variables as follows:

```yaml
dr_site_primary_url: "http://xxx.xxx.xxx:8080/ovirt-engine/api"
dr_site_primary_username: "admin@internal"
dr_site_primary_password: 123
dr_site_primary_ca_file: ca.pem

dr_site_secondary_url: "http://xxx.xxx.xxx:8080/ovirt-engine/api"
dr_site_secondary_username: "admin@internal"
dr_site_secondary_password: 123
dr_site_secondary_ca_file: ca.pem

dr_master_domain_primary_name: "data"
dr_master_domain_primary_address: "xx.xx.xx.xx"
dr_master_domain_primary_path: path
dr_master_domain_primary_dc_name: dc

dr_master_domain_secondary_name: "data"
dr_master_domain_secondary_address: "xx.xx.xx.xx"
dr_master_domain_secondary_path: path
dr_master_domain_secondary_dc_name: dc

```

I am not sure of `import_NFS` and `import_ISCSI`, should `import_NFS` take NFS in primary and add it to secondary, right? But I think there could be more such domains right?

examples/setup_recover.yml
maorlipchuk: sure, will do

roles/ovirt-clean-engine/tasks/shutdown_vms.yml
maorlipchuk: yes, fixed

roles/ovirt-clean-engine/tasks/main.yml
maorlipchuk: done

roles/ovirt-clean-engine/tasks/main.yml
maorlipchuk: done

roles/ovirt-clean-engine/tasks/main.yml
maorlipchuk: actually, since the DR handle the entire oVirt setup, there is no need to filter out any speicifc DC.
I assume that if the request will be changed then we might add this ability in the future

roles/ovirt-clean-engine/defaults/main.yml
maorlipchuk: done

roles/ovirt-clean-engine/tasks/main.yml
maorlipchuk: I've changed this TODO, we can't support export storage domain today using ovirt-ansible since export storage domain can not be attached with dirty metadata, I will open a separate bug on it

roles/ovirt-clean-engine/tasks/shutdown_vms.yml
maorlipchuk: Since I don't see a reason to wait for each VM to be stopped.
We can start the stop operation for all the VMs at once

roles/ovirt-clean-engine/tasks/shutdown_vms.yml
maorlipchuk: Yes, I suppose this comment was for run VMs

roles/ovirt-clean-engine/tasks/shutdown_vms.yml
maorlipchuk: removed

roles/ovirt-recover/defaults/main.yml
maorlipchuk: done

roles/ovirt-recover/tasks/run_vms.yml
maorlipchuk: Since I don't see a reason why to wait for each VM to run,
We can start all the VMs at once

roles/ovirt-recover/tasks/run_vms.yml
maorlipchuk: Thanks, removed

maorlipchuk:
We might have multiple master storage domains in case we have more than one DC.
I will use one list of storage domains and add a flag to indicate if it is master or not

examples/group_vars/setup_vars
```diff
@@ -0,0 +1,107 @@
+# TODO: Make hostname a variable instead of the whole API link
+dr_sites_primary_url: "http://mlipchuk.usersys.redhat.com:8080/ovirt-engine/api"
+dr_sites_primary_username: "admin@internal"
+dr_sites_primary_ca_file: "/home/mlipchuk/ovirt-share/etc/pki/ovirt-engine/ca.pem"
+
+dr_sites_secondary_url: "http://10.35.0.140:8080/ovirt-engine/api"
+dr_sites_secondary_username: "admin@internal"
+dr_sites_secondary_ca_file: "/home/mlipchuk/ovirt-share/etc/pki/ovirt-engine/ca.pem"
+
+dr_import_storage:
```
machacekondra: s/dr_import_storage/dr_import_storages

I think plural is better here, to express it's a list.

examples/group_vars/setup_vars
```diff
@@ -0,0 +1,107 @@
...
+    - {primary_name: "affinity_prod",
+       secondary_name: "affinity_recovery"}
+
+# Mapping for affinity label
+dr_affinity_label_mapping:
+    - {primary_name: "label_prod",
+       secondary_name: "label_recovery"}
+
+# Mapping for domain
+dr_domain_mapping:
```
machacekondra: dr_aaa_extensions_mapping

Would be better IMHO

examples/setup_fail_back.yml
```diff
@@ -0,0 +1,18 @@
+---
```
machacekondra: I think you wanted to name this file `setup_fall_back.yml`

examples/group_vars/setup_vars
```diff
@@ -0,0 +1,107 @@
+# TODO: Make hostname a variable instead of the whole API link
```
machacekondra: I am unsure whether it makes sense to use as an example with `group_vars`, we mostly run on `localhost` all of your roles. I think we should just show how to pass variable file in playbook

examples/group_vars/setup_vars
```diff
@@ -0,0 +1,107 @@
...
+       dr_secondary_target: ["iqn.2015-07.com.recovery.redhat:444"],
+       dr_secondary_username: "",
+       # TODO: Passwords should be encapsulated
+       dr_secondary_password: "",
+       dr_secondary_port: "",
+       dr_secondary_dc_name: "Recovery",
+       dr_secondary_name: "scsi2"}
+
+# Mapping for cluster
+dr_cluster_mapping:
```
machacekondra: I think it's better to use plural for list variables

examples/group_vars/setup_vars
```diff
@@ -0,0 +1,107 @@
...
+    - {primary_name: "label_prod",
+       secondary_name: "label_recovery"}
+
+# Mapping for domain
+dr_domain_mapping:
+    - {primary_name: "new-authz",
+       secondary_name: "internal-authz"}
+
+# Mapping for roles
+dr_role_mapping:
```
machacekondra: So currently with this design you can set for all VMs/templates the same mapping, is there a use case where you want different mapping per VM/template?

examples/group_vars/setup_vars
maorlipchuk: done

examples/group_vars/setup_vars
maorlipchuk: yes, good point

examples/group_vars/setup_vars
maorlipchuk: I will remove the TODO, but I'm not sure that I understood your comment here, do you want me to remove the group_vars?

examples/group_vars/setup_vars
maorlipchuk: done, for all of the mappings

examples/group_vars/setup_vars
maorlipchuk: For now we use it for both I think that eventually we can fetch the unregistered entity and map only the relevant properties for it, that might make it a bit more efficient

examples/group_vars/setup_vars
```diff
@@ -0,0 +1,106 @@
+dr_sites_primary_url: "http://mlipchuk.usersys.redhat.com:8080/ovirt-engine/api"
```
machacekondra: Any reason this file is in `examples/group_vars/setup_vars` ?

examples/group_vars/setup_vars
```diff
@@ -0,0 +1,106 @@
...
+dr_sites_primary_username: "admin@internal"
+dr_sites_primary_ca_file: "/home/mlipchuk/ovirt-share/etc/pki/ovirt-engine/ca.pem"
+
+dr_sites_secondary_url: "http://10.35.0.140:8080/ovirt-engine/api"
+dr_sites_secondary_username: "admin@internal"
+dr_sites_secondary_ca_file: "/home/mlipchuk/ovirt-share/etc/pki/ovirt-engine/ca.pem"
+
+dr_import_storages:
+    - {dr_domain_type: "nfs",
```
machacekondra: Just note that you don't need to use `{` and `}`. In most Ansible examples it's not used. (in whole this document)

examples/group_vars/setup_vars
```diff
@@ -0,0 +1,106 @@
...
+       # TODO: Passwords should be encapsulated
+       dr_primary_password: "",
+       dr_primary_port: "",
+       dr_primary_dc_name: "Prod",
+       dr_primary_name: "scsi2",
+
+       dr_secondary_id: "ccccf47b-ff77-4660-bf07-2809a750adb6",
+       dr_secondary_address: "10.35.0.222",
+       dr_secondary_target: ["iqn.2015-07.com.recovery.redhat:444"],
+       dr_secondary_username: "",
```
machacekondra: As this is example, if it's empty it doesn't make sense to have it here, no?

examples/setup_fail_back.yml
machacekondra: I think it's OK if you use following example:

```yaml
---
- name: Setup oVirt environment
  hosts: localhost
  connection: local
  vars:
    dr_clean: true
  vars_files:
     - disaster_recovery_passwords.yml
     - disaster_recovery_vars.yml
  roles:
    - oVirt.disaster-recovery
```

Because all other values are defauls, and if user want's to change them he can change it in `_vars` file or pass it as --extra-vars on command line.

examples/setup_recover.yml
```diff
@@ -0,0 +1,15 @@
+---
```
machacekondra: I think it's OK if you use following example:

```yaml
---
- name: Setup oVirt environment
  hosts: localhost
  connection: local
  vars_files:
     - disaster_recovery_passwords.yml
     - disaster_recovery_vars.yml
  roles:
    - oVirt.disaster-recovery
```

Because all other values are defauls, and if user want's to change them he can change it in `_vars` file or pass it as --extra-vars on command line.

roles/ovirt-clean-engine/tasks/main.yml
```diff
@@ -0,0 +1,112 @@
+- block:
+    - name: Obtain SSO token
+      ovirt_auth:
+          url: "{{ vars['dr_sites_' + dr_target_host + '_url'] }}"
+          username: "{{ vars['dr_sites_' + dr_target_host + '_username'] }}"
+          # TODO: Set password per site
+          password: "{{ password }}"
```
machacekondra: Can you fix this TODO?

machacekondra (commented):
In general it looks good IMHO.

Can we create just one role call `ovirt-disaster-recovery`?
It will have following files:
 - main.yml
 - unregister_entites.yml
- clean_engine.yml
- recover_engine.yml
- run_unregister_entites.yml

You will just move code from role to speciifc yaml files. And tag all the relevant tasks with tag like `clean-engine`, `recover-engine`, `unregister-entities`. So user can run just specific tasks with:

```bash
ansible-playbook --tags untergister-entities dr_failover.yml
```

And we will create just one role called oVirt.disaster-recovery, which will be configurable by variables and tags.

mwperina:
👍 for Ondra's proposal to have only oVirt.disaster-recovery role with specific tasks. We are already using the same approach in oVirt.infra role which be used for both creating and removing DCs/clusters/hosts/...

examples/group_vars/setup_vars
maorlipchuk: no, if you have any other suggestion for a standard location I will be happy to change it.

roles/ovirt-clean-engine/tasks/main.yml
maorlipchuk: sure, will do

examples/group_vars/setup_vars
maorlipchuk: sure, will remove

examples/group_vars/setup_vars
maorlipchuk: I think that I got an exception when I haven't declared it, but I don't remember why, I will try to remove it and update if that works

examples/setup_fail_back.yml
maorlipchuk: sure, will give it a try

examples/setup_recover.yml
maorlipchuk: sure, sounds good, will give it a try

maorlipchuk:
Great, Thanks for the help guys! I will give it a try.
I'm about to add a new play which should generate a var file using python-sdk, I guess this should be added as a tag as well.

examples/group_vars/setup_vars
machacekondra: I think it's OK to just use `examples/disaster_recovery_vars.yml`

maorlipchuk: done

maorlipchuk:
@machacekondra I eventually used 3 tags: 
- fail_back - Supports fail back scenrio from site at source map to the destination site mentioned in target_host
 - fail_over- Supports fail over scneario from site mentioned in target_host to site mentioned in source_map
 - clean_engine - Cleans the engine setup of the source_map

examples/disaster_recovery_vars.yml
```diff
@@ -0,0 +1,96 @@
+---
+dr_sites_primary_url: "http://mlipchuk.usersys.redhat.com:8080/ovirt-engine/api"
```
machacekondra, line 2: Can you please change the examples to something like:

```
dr_sites_primary_url: "https://engine1.example.com/ovirt-engine/api" 
```

examples/disaster_recovery_vars.yml
```diff
@@ -0,0 +1,96 @@
...
+dr_sites_primary_username: "admin@internal"
+dr_sites_primary_ca_file: "/home/mlipchuk/ovirt-share/etc/pki/ovirt-engine/ca.pem"
+
+dr_sites_secondary_url: "http://10.35.0.140:8080/ovirt-engine/api"
```
machacekondra, line 6: Can you please change the examples to something like:

```
dr_sites_primary_url: "https://engine2.example.com/ovirt-engine/api" 
```

examples/disaster_recovery_vars.yml
```diff
@@ -0,0 +1,96 @@
+---
+dr_sites_primary_url: "http://mlipchuk.usersys.redhat.com:8080/ovirt-engine/api"
+dr_sites_primary_username: "admin@internal"
+dr_sites_primary_ca_file: "/home/mlipchuk/ovirt-share/etc/pki/ovirt-engine/ca.pem"
```
machacekondra, line 4: /etc/pki/ovirt-engine/ca.pem

examples/disaster_recovery_vars.yml
```diff
@@ -0,0 +1,96 @@
...
+
+dr_sites_secondary_url: "http://10.35.0.140:8080/ovirt-engine/api"
+dr_sites_secondary_username: "admin@internal"
+dr_sites_secondary_ca_file: "/home/mlipchuk/ovirt-share/etc/pki/ovirt-engine/ca.pem"
```
machacekondra, line 8: /etc/pki/ovirt-engine/ca.pem

examples/disaster_recovery_vars.yml
```diff
@@ -0,0 +1,96 @@
...
+
+dr_import_storages:
+    - dr_domain_type: "nfs"
+      dr_master_domain: "True"
+      dr_primary_address: "10.35.16.43"
+      dr_primary_path: "/export/data_test7"
```
machacekondra, line 14: Maybe it would be better to add everywhere better examples of IPs/paths/names

machacekondra (commented):
@maorlipchuk Can you please move the code here:

https://github.com/oVirt/ovirt-ansible-disaster-recovery

examples/disaster_recovery_vars.yml
maorlipchuk: done

examples/disaster_recovery_vars.yml
maorlipchuk: done

examples/disaster_recovery_vars.yml
maorlipchuk: done

examples/disaster_recovery_vars.yml
maorlipchuk: done

examples/disaster_recovery_vars.yml
maorlipchuk: done

maorlipchuk:
All fixes were ported to https://github.com/oVirt/ovirt-ansible-disaster-recovery
I've added a start WIP for generate_mapping pythong code, it is not done yet.
Please let me know if I should wait with it or not

mwperina:
Closing as PR moved to https://github.com/oVirt/ovirt-ansible-disaster-recovery

PR #72, opened 2017-09-05
Add YAML file to tell CI which distros to build

ifireball:
New CI jobs move (more of) the configuration from the `jenkins` repo to the individual project repos. So now this needs to be configured here instead of in the 'Jenkins' repo.

### DIFF
--- a/automation.yaml
+++ b/automation.yaml
@@ -1,0 +1,4 @@
+distros:
+  - fc25
+  - fc26
+  - el7


automation.yaml
```diff
@@ -0,0 +1,5 @@
+distros:
+  - fc24
```
machacekondra: For master we don't need fc24 AFAIK.

ifireball: ok n/p I just added what was in the YAML in the 'jenkins' repo

ifireball:
ci build please

PR #78, opened 2017-09-12
Fix ovirt-image-template parameters

machacekondra:
Related-to: https://bugzilla.redhat.com/show_bug.cgi?id=1489454

### DIFF
--- a/roles/ovirt-image-template/tasks/qcow2_image.yml
+++ b/roles/ovirt-image-template/tasks/qcow2_image.yml
@@ -62,9 +62,8 @@
     ovirt_disks:
       auth: "{{ ovirt_auth }}"
       name: "{{ template_disk_name | default(template_name) }}"
-      interface: "{{ template_disk_format | default('virtio') }}"
-      size: "{{ template_disk_size}}"
-      format: "{{  template_disk_interface | default('cow') }}"
+      size: "{{ template_disk_size }}"
+      format: "{{ template_disk_format | default(omit) }}"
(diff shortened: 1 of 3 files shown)


roles/ovirt-image-template/tasks/qcow2_image.yml
```diff
@@ -62,9 +62,9 @@
     ovirt_disks:
       auth: "{{ ovirt_auth }}"
       name: "{{ template_disk_name | default(template_name) }}"
-      interface: "{{ template_disk_format | default('virtio') }}"
-      size: "{{ template_disk_size}}"
-      format: "{{  template_disk_interface | default('cow') }}"
+      interface: "{{ template_disk_interface }}"
+      size: "{{ template_disk_size }}"
+      format: "{{  template_disk_format }}"
```
mwperina: Default values are no longer needed?

mwperina: Ahh, please ignore, just saw them defaults/main.yml

mwperina (approved):
👍

tareqalayan (approved):

machacekondra:
@mwperina As we don't support RAW disk images, I've removed the `template_disk_format` variable.

machacekondra:
@mwperina Actually we may leave it there, but if user will use `raw` with `qcow` image it will fail. But in future we may add support to upload also `raw` images, not only `qcow2`, so we can leave the variable there, but if user will use it wrongly, it may fail for them..

Issue #77, opened 2017-09-11
ovirt-vm-infra: state == absent requires full set of profile parameters


KKoukiou:
https://github.com/oVirt/ovirt-ansible/blob/2783502690f8825d609a66e86ab2f7da41348a5d/roles/ovirt-vm-infra/tasks/main.yml#L19-L23

when trying to remove managed VMs it fails on missing parameters in `profile` dictionary.

Why is it necessary to provide this information to remove VMs ?

machacekondra:
@tareqalayan AFAIK you've added the `state` param, can you take a look?

tareqalayan:
I added the state parameter because if i don't pass it, the VM will go into running state.
While i want to create this VM from a template and the disk to be on a specific storage domain. 
So i have to create the VM and set the state to be stopped. Otherwise it will be on running state and the disk will be on the a storage that was picked randomly and not be me.

machacekondra:
Ah, ok this is actually already reported here:

 https://bugzilla.redhat.com/show_bug.cgi?id=1489893

I will fix it.

tareqalayan:
@KKoukiou  this will break us. for example for the CFME template we would like to overide the core and memory no? 

@machacekondra  can the behaviour be that if core and memory not stated it will be taken from template otherwise it will be taken from the module params?

KKoukiou:
@tareqalayan I opened issue in order to not expect all these parameters in the case that `state == absent`... 
But with a second look, I think this role is supposed to populate envs in fact, not clean them up, so I would just add it to documentation of role, that absent is not really suported, and leave behavior like this.

Regarding the default values of cores and memory, I think we can adjust our description files.

machacekondra:
In #80 I've changed the documentation and listed only states `stopped`, `present` and `running`.

machacekondra:
But it will work for you if you will not pass any `tag` and `wait_for_ip=false`.

KKoukiou:
@machacekondra I am ok with it, thanks!

PR #80, opened 2017-09-14
Vm infra fix defaults

machacekondra:
This PR fixes following:

 - remove unnecessary tags.
 - Don't require to specify VM parameters
 - when tag isn't specified, don't fail, just skip it
 - if state==stopped, don't wait for IP

Related-to: https://bugzilla.redhat.com/show_bug.cgi?id=1489893
Related-to: https://github.com/oVirt/ovirt-ansible/issues/77


machacekondra:
@mwperina

PR #82, opened 2017-09-14
Add timeouts for ovirt-vm-infra role

machacekondra:
Added new variables to manage timeouts:

 - Timeout to create single VM - `vm_infra_create_single_timeout`: 180
 - Timeout to poll creating of VM - `vm_infra_create_poll_interval`: 15
 - Timeout to wait to create all VMs - `vm_infra_create_all_timeout`: "{{ `vm_infra_create_single_timeout` * (`vms` | length) }}"
 - Timeout to retry check if VM reports IP - `vm_infra_wait_for_ip_retries`: 5
 - The delay between polling for IP report of VM - `vm_infra_wait_for_ip_delay`: 5

Related-to: https://bugzilla.redhat.com/show_bug.cgi?id=1490838

### DIFF
--- a/roles/ovirt-vm-infra/README.md
+++ b/roles/ovirt-vm-infra/README.md
@@ -12,11 +12,17 @@ Requirements
 Role Variables
 --------------
 
-| Name               | Default value     |                                              |
-|--------------------|-------------------|----------------------------------------------| 
-| vms                | UNDEF             | List of dictionaries with virtual machine specifications.   |
-| wait_for_ip        | true              | If true, the playbook should wait for the virtual machine IP reported by the guest agent.  |
-| debug_vm_create    | false             | If true, logs the tasks of the virtual machine being created. The log can contain passwords. |
+| Name                           | Default value |                                              |
+|--------------------------------|---------------|----------------------------------------------| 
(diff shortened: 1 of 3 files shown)


machacekondra:
@mwperina @pkubica

mwperina (approved):
I don't like having that many timeout/retries parameters, but I don't see any better way how to handle that

👍

machacekondra:
We can docuement just the most important one. For example we don't have to document poll intervals. We can just add `wait_for_single_vm_timeout` and compute the rest by us. But still leave an option to overwrite it.

Same for `wait_for_ip`, we can add `max_wait_time`, and compute retry and delay from it.

mwperina:
Let's have a description for all parameters that role has.

For future I'm wondering if we can somehow split parameter description into sections, for example the first section would always contain only parameters important for the flow and afterwards separate section(s) for special parameters like timeout and retries ...

machacekondra:
> For future I'm wondering if we can somehow split parameter description into sections, for example the first section would always contain only parameters important for the flow and afterwards separate section(s) for special parameters like timeout and retries ...

I can open an issue for it.

PR #83, opened 2017-09-14
Add nics to ovirt-vm-infra role

machacekondra:
This patch add possibility to manage NICs in ovirt-vm-infra role.

Related-to: https://bugzilla.redhat.com/show_bug.cgi?id=1490934


machacekondra:
@mwperina @pkubica

PR #87, opened 2017-09-18
Improve reinstalling of the hosts

machacekondra:
This patch improve the reinstalling of the host:

 - support to pass also public key instead of password
 - improve the search, so we search only for host in install_failed state and
   only hosts we specified in variables
 - document the timeouts

### DIFF
--- a/roles/ovirt-hosts/README.md
+++ b/roles/ovirt-hosts/README.md
@@ -14,15 +14,16 @@ Role Variables
 
 The `hosts` list can contain the following parameters:
 
-| Name          | Default value  | Description                           |
-|---------------|----------------|---------------------------------------|
-| name          | UNDEF          | Name of the host.                      |
-| state         | present        | Specifies whether the host is `present` or `absent`.  |
-| address       | UNDEF          | IP address or FQDN of the host.   |
-| password      | UNDEF          | The host's root password.             |
-| cluster       | UNDEF          | The cluster that the host must connect to.    |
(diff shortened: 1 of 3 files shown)


machacekondra:
@mwperina

roles/ovirt-hosts/README.md
```diff
@@ -14,15 +14,16 @@ Role Variables
...
-| cluster       | UNDEF          | The cluster that the host must connect to.    |
-| timeout       | 1200           | Maximum wait time for the host to be in an UP state.  |
-| poll_interval | 20             | Polling interval to check the host status. |
+| Name          | Default value    | Description                           |
+|---------------|------------------|---------------------------------------|
+| name          | UNDEF (Required) | Name of the host.                      |
+| state         | present          | Specifies whether the host is `present` or `absent`.  |
+| address       | UNDEF (Required) | IP address or FQDN of the host.   |
+| password      | UNDEF            | The host's root password. Required if <i>public_key</i> is false. |
+| public_key    | UNDEF            | If <i>true</i> the public key should be used to authenticate to host. |
```
mwperina, line 23: I thought that this param will contain the public key of the host. So why boolean?

machacekondra: No, it's true/false, we don't support specify custom public key, only the default one - the one deployed by host-deploy proccess.

mwperina: Ahh, right

mwperina (approved):
👍

Issue #48, opened 2017-08-21
ovirt-aaa-jdbc: allow setting attributes via this role 


tareqalayan:
@mwperina 
@machacekondra 

I would like to have the ability to set some attributes from: 
--attribute=[<name>=<value>]
    Available names:
      department
      description
      displayName
      email
      firstName
      lastName
      title

the command should be something like 
"/usr/bin/ovirt-aaa-jdbc-tool user {{ (item.state is undefined or item.state == 'present') | ternary('add','delete') }} {{ item.name }} --attribute=firstName={{ item.name }} --attribute=department='My department'"

I am thinking to add an extra_ops parameter:
    users:
     - name: user1
       authz_name: internal-authz
       password: 1234568
       valid_to: "2018-01-01 00:00:00Z"
       extra_ops: --attribute=firstName=bob --attribute=department='My department'"
     - name: user2
       authz_name: internal-authz
       password: 1234568
       valid_to: "2018-01-01 00:00:00Z"

would this be fine by you?
or do you have another suggestion?

machacekondra:
why not just have `attributes` dicitnary?

```yaml
users:
     - name: user1
       authz_name: internal-authz
       password: 1234568
       valid_to: "2018-01-01 00:00:00Z"
       attributes:
         firstname: 'bob'
         department: 'My department'
```

PR #88, opened 2017-09-18
ovirt-aaa-jdbc: allow set attributes for user

tareqalayan:
When you add a user via ovirt-aaa-jdbc you can set attributes with
```
--attribute=[<name>=<value>]
    Available names:
      department
      description
      displayName
      email
      firstName
      lastName
      title
```

Example of how to send these variables in a playbook:
```yaml
vars:
  users:
    - name: user1
      authz_name: internal-authz
      password: 1234568
      valid_to: "2018-01-01 00:00:00Z"
    - name: user2
      authz_name: internal-authz
      password: 1234568
      valid_to: "2018-01-01 00:00:00Z"
      attributes:
        firstName: 'alice'
        department: 'Quality Engineering'
```

Fixes: https://github.com/oVirt/ovirt-ansible/issues/48

### DIFF
--- a/roles/ovirt-aaa-jdbc/tasks/main.yml
+++ b/roles/ovirt-aaa-jdbc/tasks/main.yml
@@ -16,6 +16,21 @@
     - ovirt-aaa-jdbc
     - users
 
+- name: Update users according to attributes
+  command: "{{ aaa_jdbc_prefix }}/ovirt-aaa-jdbc-tool user edit {{ item.name }} {% for attr, value in item['attributes'].iteritems() %} --attribute={{ attr }}='{{ value }}' {% endfor %}"
+  with_items:
+    - "{{ users | default([]) }}"
+  register: out_users
+  when: "item.attributes is defined"
+  changed_when: "out_users.rc != 5 and out_users.rc != 4"
+  # 5 == user already exists
+  # 4 == no user to be removed
+  # 0 == all OK
+  failed_when: "out_users.rc != 5 and out_users.rc != 0 and out_users.rc != 4"
+  tags:
+    - ovirt-aaa-jdbc
+    - users
+
 # FIXME: when user try to change the password which was already set in history
 # but is not current password we continue with changed=false:
 - name: Manage internal users passwords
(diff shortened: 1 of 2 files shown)


roles/ovirt-aaa-jdbc/tasks/main.yml
```diff
@@ -3,7 +3,7 @@
 ## User & groups internal
 ################################
 - name: Manage internal users
-  command: "{{ aaa_jdbc_prefix }}/ovirt-aaa-jdbc-tool user {{ (item.state is undefined or item.state == 'present') | ternary('add','delete') }} {{ item.name }}"
+  command: "{{ aaa_jdbc_prefix }}/ovirt-aaa-jdbc-tool user {{ (item.state is undefined or item.state == 'present') | ternary('add','delete') }} {{ item.name }} {% if item['attributes'] is defined %} {% for attr, value in item['attributes'].iteritems() %} --attribute={{ attr }}={{ value }} {% endfor %} {% endif %}"
```
petr-balogh: {{ value }} maybe value should be surrounded by quotas to allow values with spaces?

machacekondra: Maybe better to have:

```
$ ovirt-aaa-jdbc-tool user edit user1 {% if item['attributes'] is defined %} {% for attr, value in item['attributes'].iteritems() %} --attribute={{ attr }}={{ value }} {% endfor %} {% endif %}"
```

That way the attributes will be always changed. Not only when user is added.

tareqalayan: will fix.

tareqalayan: @machacekondra  edit will also add user if it is not exist?

machacekondra: @tareqalayan no, it should be another task

tareqalayan: @machacekondra  will fix. thanks

roles/ovirt-aaa-jdbc/tasks/main.yml
```diff
@@ -16,6 +16,20 @@
     - ovirt-aaa-jdbc
     - users
 
+- name: Update users according to attributes
+  command: "{{ aaa_jdbc_prefix }}/ovirt-aaa-jdbc-tool user edit {{ item.name }} {% if item['attributes'] is defined %} {% for attr, value in item['attributes'].iteritems() %} --attribute={{ attr }}='{{ value }}' {% endfor %} {% endif %}"
```
machacekondra:
```
when: "item.attributes is defined"
```

would be imo more readable.

tareqalayan: I agree, will fix.

machacekondra (approved):
Thanks.

PR #89, opened 2017-09-19
Fix the flow of the ovirt-vm-infra role

machacekondra:
This patch changes the flow of the tasks in the ovirt-vm-infra role. We
need to first create a stopped VM, then attache the needed disks/nics,
and later start the VM if needed.
    
If VM is already started prior to this role, we try to hotplug disks and
nics.
    
Related-to: https://bugzilla.redhat.com/show_bug.cgi?id=1490930


machacekondra:
@mwperina

PR #92, opened 2017-09-20
Add suffix to modules with bug fixes

machacekondra:
This patch add a suffix internal_{ansible_version} to all modules in
library. The {ansible_version} indicates which version of ansible
provides the fix, which we need to ship with the role.

So if Ansible {ansible_version} is released we can remove the module
from library and rename the task to the offical upstream module.

Related-to: https://bugzilla.redhat.com/show_bug.cgi?id=1487082


machacekondra:
@mwperina

roles/ovirt-vm-infra/tasks/create_vms.yml
```diff
@@ -4,7 +4,9 @@
     pattern: "name={{ current_vm.name }}"
 
 - name: "Create VM {{current_vm.name }}"
-  ovirt_vms:
+  # TODO: Change when Ansible 2.4 is released:
+  # ovirt_vms:
+  ovirt_vms_24:
```
mwperina, line 9: Why not ovirt_vms_internal_24?

machacekondra: Ah, no reason, forget to add. It's not an issue, because the module is named `ovirt_vms_24`. I've tested and it was executed successfully.

mwperina: Sure, just asked :-)

PR #93, opened 2017-09-20
Rebase ovirt_vms module with Ansible fix #27382

machacekondra:
This patch rebase ovirt_vms module with Ansible fix #27382.

Related-to: https://bugzilla.redhat.com/show_bug.cgi?id=1491010


machacekondra:
@mwperina

PR #95, opened 2017-09-21
Skip ovn role if ovn_central var isn't specified

machacekondra:
In case user don't specify `host_deploy_ovn_central` or `ovn_central` we need to skip whole ovn role rather then fail, that it's not specified.


machacekondra:
@dominikholler

machacekondra:
ci build please

dominikholler (approved):
Looks very good.

PR #106, opened 2017-10-06
Add host-upgrade role

machacekondra:
We update and check of the update of the host using single playbook called `ovirt-host-upgrade.yml`. This playbooks just executes the `ovirt-host-upgrade` role. The difference between check-updated and update is following:

__Check-update__
 1. Run playbook with dry-mode (--check) and env variable `ANSIBLE_STDOUT_CALLBACK=hostupgradeplugin`.
 2. If any tasks tagged by `updatecheck` reports that there need to be installed/updated package `roles/ovirt-host-upgrade/callback_plugins/hostupgradeplugin.py` we save those packages and return list of all those packages.
 3. At the engine side we parse the stdout and save all packages that needs update.

__Update__
 1. Run playbook without dry-mode (--check) and default stdout callback plugin - so log stdout to ovirt-host-mgmt-ansible-*.log file.

Related-to: https://bugzilla.redhat.com/show_bug.cgi?id=1380498

### DIFF
--- a/playbooks/ovirt-host-upgrade.yml
+++ b/playbooks/ovirt-host-upgrade.yml
@@ -1,0 +1,6 @@
+- hosts: all
+  remote_user: root
+  gather_facts: false
+
+  roles:
+    - ovirt-host-upgrade
--- a/roles/ovirt-host-upgrade/README.md
+++ b/roles/ovirt-host-upgrade/README.md
@@ -1,0 +1,37 @@
+oVirt Host Upgrade
+==================
+
+The `ovirt-host-upgrade` role updates all packages on the host and installs ovirt-host package if it isn't installed.
+
+Requirements
+------------
+
+ * Ansible version 2.0
+
+Role Variables
+--------------
+
+No.
+
+Dependencies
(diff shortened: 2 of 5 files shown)


machacekondra:
@mwperina

mwperina (commented):
Generally looks good to me, but one question came to my mind: upgrading all packages on EL/Fedora hosts is the right way, but what about NGN? Should we also upgrade all packages for NGN or only ovirt-node-ng-image-update as in previous versions?

@masayag @evol262

mwperina:
Ryan/Sandro any opinion on upgrading all packages on NGN or just ovirt-node-ng-image-update?

@evol262 @sandrobonazzola

ghost:
We actually restrict the available packages on NGN in /etc/yum.repos.d, with most repos disabled.

"yum -y upgrade" should only catch image-update and other packages which *should* be updated on NGN anyway, so it's safe

mwperina (approved):
OK, let's keep it simple and make no difference between EL host and NGN

Issue #85, opened 2017-09-15
Add rest of `ovirt_clusters` module parameter to ovirt-clusters role


machacekondra:
There is a lot of parameters missing from:
http://docs.ansible.com/ansible/latest/ovirt_cluster_module.html

PR #111, opened 2017-10-11
ovirt-cluster: Add all parameters from upstream module

machacekondra:
This patch add all attributes which are supported by upstream
ovirt_clusters module:
 http://docs.ansible.com/ansible/latest/ovirt_cluster_module.html

Fixes: https://github.com/oVirt/ovirt-ansible/issues/85


machacekondra:
ci build please

PR #112, opened 2017-10-13
Template Image Resize

cwilkers:
Add ability to resize template image if smaller than intended.
- Adds requirement for qemu-img, but checks if binary is present to avoid failure
- Reads current size of qcow2 image

VMs created with resized image may automatically expand to use space. (rhel 7.4 kvm tested)

Functionality needed for OpenShift 3.6 install.

@machacekondra please review


ovirt-infra:
Hello contributor, thanks for submitting a PR for an oVirt project! 

I am the bot who triggers CI builds on the oVirt Jenkins server.
As a security measure I will not run automated tests on PRs that are not from white-listed contributors.

In order to allow automated tests to run, please ask one of the project maintainers to review the code and then do one of the following:

1. Type `ci test please` on this PR to trigger automated tests for it.
2. Type `ci add to whitelist` on this PR to trigger automated tests for it and also add you to the contributor white-list so that your future PRs will be tested automatically.
3. If you are planning to contribute to more than one project, maybe it's better to ask them to add you to the project organization, so you'll be able to run tests for all the organisation's projects.

Note: You might see multiple messages like this showing up on a single PR, this is due to a known bug in the GitHub trigger plugin on Jenkins, the oVirt infra team is working on fixing this via new testing flow in the near future, so we'll be able to post a single message per PR.

roles/ovirt-image-template/tasks/qcow2_image.yml
```diff
@@ -22,6 +22,27 @@
   tags:
     - ovirt-template-image
 
+- name: Check for qemu-img
+  stat: path=/usr/bin/qemu-img
```
machacekondra: Hmm, this way we need to require `qemu-img` package. I think it's not shipped in channel on engine.
Maybe it would be better to create a template with disk which has size same as qcow disk image, and then resize it via oVirt API.

Then write a small script which find the size of qcow image, taking a look at[1], shouldn't be that hard.

[1] https://people.gnome.org/~markmc/qcow-image-format.html

cwilkers: That is possible, I can read the size parameter directly with the following snippet:

    hexdump -s 24 -l 8 /tmp/ovirt_image_data

Is it the intent of the ovirt-ansible project to always be run (or able to be run) from the engine?

machacekondra: In general it should be possible to run it on any machine. But we need to say what are the requirements. The current patch is OK, if we say that user have to run the role on the machine with installed `qemu-img`. But I think it's better to make it possible to run it even on clean el7 machine, if it's possible.

cwilkers: I'm looking at other options, but one thing I'd like to add in is that `hexdump` is a bit clunky to script, and that the `xxd` tool from `vim-common` provides better output AND includes the ability to write to a binary file at an offset and length should we want to edit the file before uploading without `qemu-img`.

`vim-common` is also missing from an engine's standard install, (and the RHEL 7.4 KVM guest image) but is closer to a RHEL standard.

machacekondra: I think writing a small python script would be the best approach, what do you think?

cwilkers: How about this for a beginning?

```
#!/usr/bin/python

import sys
import struct

if __name__ == "__main__":
    try:
        target_gib = int(sys.argv[1])
        qcow_filename = sys.argv[2]
    except IndexError:
        print "Usage: " + sys.argv[0] + " <size in GiB> <filename>"
        sys.exit(1)
    with file(qcow_filename, 'r+b') as fh:
        fh.seek(24,0)
        current_packed = fh.read(8)
        (size_file,) = struct.unpack('>Q', current_packed)
        target_bytes = target_gib * 2**30
        if target_bytes > size_file:
            newsize_packed = struct.pack('>Q', target_bytes)
            fh.seek(24,0)
            fh.write(newsize_packed)
```

machacekondra: I just wonder whether is it safe to rewrite the size of the qcow image like that.
It's doing quite a lot of stuff:

 - https://github.com/qemu/qemu/blob/master/qemu-img.c#L3415
 - https://github.com/qemu/qemu/blob/530049bc1dcc24c1178a29d99ca08b6dd08413e0/block.c#L3556

I think it would be better to just get the size of the qcow image, and create a template with that image. Then check if `template_disk_size` is bigger and if yes call `ovirt_disks` module to handle resize.

Also I've just noticed that file[1] command is parsing the qcow files as well:

```bash
$ file cirros-0.3.0-x86_64-disk.img 
cirros-0.3.0-x86_64-disk.img: QEMU QCOW Image (v2), 41126400 bytes

$ file cirros-0.3.0-x86_64-disk.img | awk '{print $(NF-1)}'
41126400

```

[1] https://github.com/file/file/blame/a25f230f1e7adf1632592d5c81a2d6cc86559c67/magic/Magdir/virtual#L33

machacekondra: calling the `ovirt_disks` should be just a matter of:

```yaml
ovirt_disks:
  id: "{{ ovirt_disk.id }}"
  auth: "{{ ovirt_auth }}"
  size: "{{ template_disk_size }}"
when: template_disk_size > qcow_disk_size_in_gb
```

cwilkers: A diff of a `qemu-img` re-sized ovirt_image_data and one my python script re-sized comes up with no differences. All the complexity in that snippet of code seems to be around the truncate case, which we're purposely avoiding.

I can test with the api, but my initial problem was that re-sizing the template resulted in a correctly sized disk image on the RHV server (even all VMs are created with 60G root disks) but the OS does not see the extra size because the image is not getting adjusted, only the backing store.

machacekondra: OK, I just wasn't sure if just re-writing that value is enough, if that's enough, let's use your script.
Thanks a lot!

mykaul:
Can anyone from the storage team review this?

mwperina:
Allon, can someone from storage team review this PR?

@mureinik

roles/ovirt-image-template/tasks/qcow2_image.yml
cwilkers: Additional testing has shown that my test image received more changes than just the size header when qemu-img re-sized it. If I run my script against a newly downloaded qcow2, it makes the file unreadable by qemu-img.

cwilkers: I've also tested with the `ovirt_disks` method outlined earlier, and it does not affect the disk size. (The temporary VM disk remains at 10GiB despite attempts to re-size it to 60GiB.)

I also attempted to set the size parameter on template VM creation, also to no effect.

mureinik:
@cwilkers  I seem to be missing something basic here - why aren't you just creating the disk with the right size to begin with?
Can you describe the full flow of what you're trying to achieve here?

machacekondra:
@mureinik If you have qcow image which has let's say have 1GiB space. And you create a disk in oVirt which has 5GiB and upload there the 1GiB qcow image, `fdisk` reports that the disk has only 1GiB of space, not 5GiB.

The way which works is just create disk which has same space as the qcow disk, and then resize using API.

mureinik:
@machacekondra `fdisk` from within the guest? from the host? Sounds like a misuse of the API - but again, I think we have a breakdown in communication here. Can you give an example usecase and explain what exactly you're trying to achieve?

machacekondra:
@mureinik within guest.

Let's say I have locally download this qcow image:

 https://download.cirros-cloud.net/0.3.5/cirros-0.3.5-x86_64-disk.img

When I execute info on this disk:
```bash
$ qemu-img info cirros.qcow2

image: cirros.qcow2
file format: qcow2
virtual size: 39M (41126400 bytes)
disk size: 13M
cluster_size: 65536
Format specific information:
    compat: 0.10
    refcount bits: 16
```

I can see the image has `41126400` bytes. Now I want to upload this image to oVirt via API.
I use following script:

 https://github.com/oVirt/ovirt-engine-sdk/blob/master/sdk/examples/upload_disk.py

And so the disk in oVirt shows 10GiB of space. When I create a VM from this disk, and I boot the VM, log-in to console, log-in to system, I run fdisk and I see the disk has just 41126400bytes of space, not 10GiB.

That's the whole point of this patch.

mureinik:
@machacekondra one of us is missing something (quite possibly me) - `qemu-img resize` will just resize the size of the file. It will not change the partitions the guest sees (e.g., via `fdisk`). I don't see any point in resizing the image post upload instead of just creating it properly (using the proper `provisioned_size` and `initial_size` parameters in the script you provided. Once you have a qcow file that's large enough, you can use `fdisk` from the guest to resize the partitions (see, e.g. https://gist.github.com/larsks/3933980)

machacekondra:
@mureinik Yeah, I know. When you create disk which has same size has qcow2 image you want to upload, and later resize the disk to size you really need all works just good. That's what we want to do in this patch.

The problem is when you create disk which is larger then the qcow2 image you want to upload.
So let's say my qcow image has 1GiB, but I want to upload it to disk which has 5GiB. The upload proccess succed, but the image has usable only 1GiB, not 5GiB.

If I run `qemu-img` command on storage:
```bash
$ qemu-img info /data/c2c612b7-6470-48aa-8b6d-7f77ca32e1b8/images/d360fe54-5640-4db1-b49e-79d182a21a50/7f14c587-b6b0-4678-88c5-be23b9d80acc
image: /data/c2c612b7-6470-48aa-8b6d-7f77ca32e1b8/images/d360fe54-5640-4db1-b49e-79d182a21a50/7f14c587-b6b0-4678-88c5-be23b9d80acc
file format: qcow2
virtual size: 39M (41126400 bytes)
disk size: 21M
cluster_size: 65536
Format specific information:
    compat: 1.1
    lazy refcounts: false
    refcount bits: 16
    corrupt: false
```

If I check engine database:
```
engine=# select size from all_disks where image_guid = '7f14c587-b6b0-4678-88c5-be23b9d80acc';
    size    
------------
 5368709120
```

machacekondra:
@cwilkers I've submited new commit with `ovirt_disk` module used to resize, worked for me just well, can you check and try if possible?

roles/ovirt-image-template/tasks/qcow2_image.yml
```diff
@@ -112,6 +102,27 @@
     tags:
       - ovirt-template-image
 
+  - block:
+    - name: Resize disk
```
cwilkers: @machacekondra Ansible-playbook is generating an error here:
```
$ ansible-playbook -e@test/ocp-vars.yaml playbooks/ovirt-vm-infra.yaml 
ERROR! no action detected in task. This often indicates a misspelled module name, or incorrect module path.

The error appears to have been in '/home/cwilkers/ovirt-ansible/roles/ovirt-image-template/tasks/qcow2_image.yml': line 106, column 7, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

  - block:
    - name: Resize disk
      ^ here


```

machacekondra: You have to use Ansible 2.4, because `ovirt_disks` was renamed to `ovirt_disk`. `ovirt_disks` can still be used but it generated warning in 2.4 and will be removed in 2.8

cwilkers: That's a problem, as I'm trying to use GA RHEL code to perform the reference architecture... I have 2.3.2.0.

We can go ahead with the ovirt_disks for backwards compatibility, but if this module now requires Ansible >= 2.4, that may cause some headaches down stream.

machacekondra: This will go for RHV/oVirt 4.2, where Ansible 2.4 will be for sure. If I will backport also to 4.1.z, I will need to change the name.

cwilkers:
@mureinik To answer the question of use case, I'm updating a reference architecture where OpenShift is installed on RHV in a highly automated fashion. For OpenShift 3.5, we pulled RHEL 7.3 guest KVM images from the customer portal, uploaded those to RHV using the ovirt-ansible code to create a template, and install eight VMs from that template to make up the OpenShift cluster.
For OpenShift 3.6, the 10GiB default image size is insufficient (checks in 3.6 require /var to have >= 15GiB)
The nice thing about the KVM guest images is that they will resize on first boot to fill an available volume, so if the qcow is resized properly to, say, 60GiB before VM creation, the VM template will have a 60GiB vda.

I can do this with qemu-img before uploading to RHV, and it "just works".

cwilkers:
@machacekondra I renamed the ovirt_disk in your code to ovirt_disks and was able to get everything working!

In my own experiment, I was trying to resize the disk right after upload, before vm creation, and I also did not have a check/wait for the successful resize.

mykaul:
Perhaps the /var should be on a separate disk?

cwilkers:
@mykaul That would make the most sense if we were installing via kickstart.

roles/ovirt-image-template/tasks/qcow2_image.yml
```diff
@@ -86,11 +90,32 @@
         - id: "{{ ovirt_disk.id }}"
           bootable: true
           interface: "{{ template_disk_interface }}"
+          size: "{{ template_disk_size }}"
       nics: "{{ template_nics }}"
     when: ovirt_templates | length == 0
     tags:
       - ovirt-template-image
 
+
```
machacekondra: Shouldn't be a `block` here?

roles/ovirt-image-template/files/qcow2_resize.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/python
```
machacekondra: Can you remove this file? I think it's not needed.

roles/ovirt-image-template/tasks/qcow2_image.yml
```diff
@@ -86,11 +90,32 @@
         - id: "{{ ovirt_disk.id }}"
           bootable: true
           interface: "{{ template_disk_interface }}"
+          size: "{{ template_disk_size }}"
```
machacekondra: Can you remove this line? It's not needed.

machacekondra (changes requested):

cwilkers:
@machacekondra 
Thanks, I have removed the unnecessary python script, the size parameter from the template VM instantiation, and adjusted indentation of the `when` statement so it properly matches `block`.

machacekondra (approved):

PR #122, opened 2017-10-26
role to install packages

tareqalayan:
required for engine-deployment role

### DIFF
--- a/roles/ovirt-engine-install-packages/README.md
+++ b/roles/ovirt-engine-install-packages/README.md
@@ -1,0 +1,40 @@
+oVirt Engine Install Packages
+======================
+
+This role installs all necessary packages for oVirt engine deployment.
+
+Target Systems
+--------------
+
+* engine
+
+
+Requirements
+------------
+
+Preinstalled clean environment with configured repositories.
+
+
+Role Variables
+--------------
+
+| Name                                        | Default value                                                         |                                                  |
+|---------------------------------------------|-----------------------------------------------------------------------|--------------------------------------------------|
+| ovirt_engine_install_packages_package_list  | [ovirt-engine,ovirt-engine-dwh-setup,ovirt-engine-extension-aaa-ldap] |  List of the packages to be installed on engine. |
+
+
+Dependencies
+------------
+
+* ovirt-repositories
+
+
+Example Playbook
+----------------
+
+```yaml
+---
+- hosts: engine
+  roles:
+    - ovirt-engine-install-packages
+```
--- a/roles/ovirt-engine-install-packages/defaults/main.yml
+++ b/roles/ovirt-engine-install-packages/defaults/main.yml
@@ -1,0 +1,6 @@
+---
+# default vars for ovirt-engine-install-packages
+ovirt_engine_install_packages_package_list:
+  - ovirt-engine
+  - ovirt-engine-dwh-setup
+  - ovirt-engine-extension-aaa-ldap
--- a/roles/ovirt-engine-install-packages/meta/main.yml
+++ b/roles/ovirt-engine-install-packages/meta/main.yml
@@ -1,0 +1,19 @@
+---
+galaxy_info:
+  author: Petr Kubica
+  description: Role to install packages required for deploying oVirt engine.
+  company: Red Hat, Inc.
+
+  license: Apache License 2.0
+
+  min_ansible_version: 2.3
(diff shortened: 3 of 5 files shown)


roles/ovirt-engine-install-packages/meta/main.yml
```diff
@@ -0,0 +1,22 @@
+---
+galaxy_info:
+  author: Tareq Alayan
+  description: Role to install packages required for deployin oVirt engine.
```
machacekondra: deploying

roles/ovirt-engine-install-packages/meta/main.yml
```diff
@@ -0,0 +1,22 @@
...
+
+  license: Apache License 2.0
+
+  min_ansible_version: 2.3
+
+  platforms:
+  - name: EL
+    versions:
+    - 7
+  - name: Fedora
```
machacekondra, line 15: Let's not explicitly set fedora versions for all roles, so we don't need to maintain it

roles/ovirt-engine-install-packages/defaults/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+# default vars for ovirt-ovirt-engine-install-packages-packages role
+ovirt_engine_dwh: True
+ovirt_engine_version: 4.1
+ovirt-engine-extension-aaa-ldap: True
```
machacekondra: What about aaa-misc, logger etc? I just wonder whether it wouldn't be better to set a list of additional packages

roles/ovirt-engine-install-packages/meta/main.yml
tareqalayan: done

roles/ovirt-engine-install-packages/meta/main.yml
tareqalayan: ok

roles/ovirt-engine-install-packages/defaults/main.yml
tareqalayan: changing to list.

roles/ovirt-engine-install-packages/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+Requirements
+------------
+
+Preinstalled clean environment with configured repositories.
+
+Role Variables
+--------------
+
+None
```
machacekondra: You miss the documentation of variables here.

roles/ovirt-engine-install-packages/defaults/main.yml
```diff
@@ -0,0 +1,10 @@
+---
+# default vars for ovirt-ovirt-engine-install-packages
+ovirt_engine_version: 4.1
+ovirt_engine_type: ovirt-engine
+packages:
```
machacekondra: please use `ovirt_engine_` prefix

roles/ovirt-engine-install-packages/tasks/main.yml
```diff
@@ -0,0 +1,17 @@
...
+    state: installed
+    update_cache: yes
+  tags:
+    - skip_yum_install_ovirt_engine
+
+- name: yum install packages
+  yum:
+    name: "{{ item }}"
+    state: present
+  with_items: "{{ packages | default([]) }}"
```
machacekondra: the `| default([])` is not needed as you have it set to `ovirt-engine-extensions-aaa-ldap`

roles/ovirt-engine-install-packages/tasks/main.yml
```diff
@@ -0,0 +1,17 @@
+---
+# tasks file for ovirt-engine-install-packages
+- name: yum install engine
+  yum:
+    name: "{{ ovirt_engine_type }}"
```
machacekondra: I think the name is ovirt-engine also for downstream , no?

KKoukiou:
@machacekondra @mwperina wdyt?

machacekondra:
I've already mentioned somewhere that this should not be a role, but this should be part of setup role.

KKoukiou:
@machacekondra I agree actually. I probably missed this comment. So, closing this PR.

KKoukiou:
I don't have permission. It's Tareqs. @tareqalayan please close this PR, see reason in comments above.

Issue #117, opened 2017-10-23
Start/stop pinned VMs before/after the upgrade of the host, not of the whole cluster


machacekondra:
Currently we stop pinned VMs before the cluster upgrade not the specific host upgrade, where the VM is pinned. We should just stop the VM when we are upgrading the host not the cluster, so the VM is quickly up.

mykaul:
As well as bring those VMs up when the host is back up, and not at the end of the whole upgrade process.

machacekondra:
@mykaul The problem is, that during cluster upgrade we use 'ClusterInMaintenance' policy, so only HA VMs can be started. So if VM won't be HA we can start it at the end, is that OK?

machacekondra:
cc @mwperina

mykaul:
It's better than what we have today. It's not optimal, but I think (in 4.2?) you can have a pinned VM set as HA too.

machacekondra:
Moved to: https://github.com/oVirt/ovirt-ansible-cluster-upgrade/issues/9

PR #126, opened 2017-10-31
ovirt_cluster_upgrade: Stop pinned VMs before host upgrade

machacekondra:
Fixes: https://github.com/oVirt/ovirt-ansible/issues/117


machacekondra:
@mwperina

roles/ovirt-cluster-upgrade/tasks/main.yml
```diff
@@ -83,14 +78,11 @@
       with_items:
         - "{{ stopped_vms | default([]) }}"
   
-    - include: pinned_vms.yml
-  
     # Update only those hosts that aren't in list of hosts were VMs are pinned
     # or if stop_non_migratable_vms is enabled, which means we stop pinned VMs
```
mwperina, line 82: Hmm, this comment is not valid anymore, right?

roles/ovirt-cluster-upgrade/tasks/main.yml
```diff
@@ -101,22 +93,22 @@
...
         ovirt_vms:
           auth: "{{ ovirt_auth }}"
           name: "{{ item }}"
+          state: running
         ignore_errors: yes
         with_items:
-          - "{{ stopped_vms | default([]) }}"
-  
-      - name: Start again pin to host VMs
+          - "{{ pinned_vms_down }}"
```
mwperina, line 103: Are we starting pinned-to-host VMs when upgrade of the host is finished or after upgrade of all hosts finished?

machacekondra: Of all host is finished, because we can't start any VM, because of `cluster_maintanence` cluster scheduling policy.

roles/ovirt-cluster-upgrade/tasks/main.yml
machacekondra: will fix

roles/ovirt-cluster-upgrade/tasks/main.yml
mwperina: Hmm, I remember Yaniv talked about this requirement, that we should restart such VMs as soon as possible (for example if OpenShift is running on top of oVirt and we need to shutdown one of master node VM during host upgrade we should start that VM again right after host is successfully upgraded). Or did I misunderstood?

@mykaul @mgoldboi

mykaul: This is indeed sub-optimal. The use case I know of is multiple OpenShift masters, each pinned to a specific host. So once the first host is upgraded, you want to re-run the master pinned to it, so you can safely proceed to the next host.

In essence, this is going to be the situation in every VM level HA, where you pin the HA VMs to specific hosts.

machacekondra: I've talked about this issue with Martin Sivak and he told me that first they should use affinity rather then pinning, so we should add support for affinity at first place.

Then, the `cluster_maintainence` policy is not so important to be enabled, so we can workaround this issue, that we don't enable that policy in case user use pinned non-migratable VMs.

machacekondra:
Issue is tracked in new repo ovirt-ansible-cluster-upgrade:

 https://github.com/oVirt/ovirt-ansible-cluster-upgrade/issues/9

PR #128, opened 2017-11-02
ovirt-host-deploy-firewalld: Fix ports per cluster service

mwperina:
* Sorts services according to their names in vars files
* Moves incorrectly placed service to match original iptables rules:
  - rpc-bind is relevant only for gluster
  - vdsm and libvirt-tls are required for both virt and gluster
* Adds missing gluster swift port to gluster


mwperina:
@machacekondra

machacekondra (approved):

PR #129, opened 2017-11-02
ovirt-host-upgrade: update cache

machacekondra:
This patch adding a cache update before running check for upgrade or actual upgrade.


machacekondra:
@mwperina

mwperina (approved):

PR #130, opened 2017-11-03
Added playbook to cleanup after removing hosted engine host.

voglovaMiroslava:


mwperina (commented):
Looks good to me, but please post also engine side patch so we can review together

@sandrobonazzola @marsik

mwperina:
@tiraboschi

playbooks/ovirt-host-remove.yml
```diff
@@ -0,0 +1,5 @@
+- hosts: all
+  remote_user: root
+
+  roles:
+    - ovirt-host-remove
```
MarSik: No newline at the end of file..

roles/ovirt-host-remove/tasks/main.yml
```diff
@@ -0,0 +1,27 @@
+---
+- name: Stops the ovirt-ha-agent service
+  service:
+    name: ovirt-ha-agent
+    state: stopped
+    enabled: no
+  when: host_remove_is_hosted_engine
```
MarSik: Where is the fact coming from? Does this really have to be conditional?

voglovaMiroslava: Before engine calls this ansible playbook, it sets this fact.
Service probably does not have to be conditional, not sure what it does if ovirt-ha-agent does not exists.
Copy & file needs to be conditional. If it is not hosted engine host then this role will fail, which will create warning in logs that does not have to be there.

MarSik: Why not check the presence of the file in the playbook itself then? I am not a big fan of external facts..

MarSik: or use ignore_error: yes

voglovaMiroslava: Ok, will do.

machacekondra (approved):

playbooks/ovirt-host-remove.yml
```diff
@@ -0,0 +1,37 @@
+- hosts: all
+  remote_user: root
+  vars:
+      hosted_engine_config_file: "/etc/ovirt-hosted-engine/hosted-engine.conf"
+
+  tasks:
+    - name: Checks hosted engine configuration file
+      stat:
+        path: "{{ hosted_engine_config_file }}"
+      register: conf_file
```
mwperina, line 10: We are checking existence of a file several times, so the playbook may be inconsistent if some other process creates/removes the file during execution. Wouldn't it be better to check the file existence only once and afterward use only the result?

mwperina (changes requested):

playbooks/ovirt-host-remove.yml
voglovaMiroslava: I thought _register_ assigns result of command (in this case stat) to the variable (conf_file). So, if I understand it correctly, the check should be done only once and then we work only with result of this check.

mwperina: Right, my mistake, thanks!

mwperina (approved):

PR #131, opened 2017-11-07
Add role and playbook for packing an OVA

ahadas:
This is a WIP, opening a PR to get some preliminary feedback

### DIFF
--- a/roles/ovirt-ova-export-pre-pack/tasks/main.yml
+++ b/roles/ovirt-ova-export-pre-pack/tasks/main.yml
@@ -1,31 +1,31 @@
 ---
-    - name: Examine target directory
-      stat:
-        path: "{{ target_directory }}"
-      register: target_directory_stats
+- name: Examine target directory
+  stat:
+    path: "{{ target_directory }}"
+  register: target_directory_stats
 
-    - name: Fail if target directory does not exist
-      fail: msg="Target directory does not exist"
-      when: not target_directory_stats.stat.exists
+- name: Fail if target directory does not exist
+  fail: msg="Target directory does not exist"
+  when: not target_directory_stats.stat.exists
 
-    - name: Fail if target directory is not a directory
-      fail: msg="Target directory is not a directory"
-      when: not target_directory_stats.stat.isdir
+- name: Fail if target directory is not a directory
+  fail: msg="Target directory is not a directory"
+  when: not target_directory_stats.stat.isdir
 
-    - name: Fail if target directory is not writeable
-      fail: msg="Target directory is not writeable"
-      when: not target_directory_stats.stat.writeable
+- name: Fail if target directory is not writeable
+  fail: msg="Target directory is not writeable"
+  when: not target_directory_stats.stat.writeable
 
-    - name: Removing the temporary file 
-      vars:
-        path: "{{ target_directory }}/{{ ova_name }}.tmp"
-      file: path="{{ path }}" state=absent
-      when: validate_only is not defined
+- name: Removing the temporary file
+  vars:
+    path: "{{ target_directory }}/{{ ova_name }}.tmp"
+  file: path="{{ path }}" state=absent
+  when: validate_only is not defined
 
-    - name: Prepare temporary path for the OVA file
-      vars:
-        path: "{{ target_directory }}/{{ ova_name }}.tmp"
-      file: path="{{ path }}" state=touch
-      register: ova_file
-      when: validate_only is not defined
+- name: Prepare temporary path for the OVA file
+  vars:
+    path: "{{ target_directory }}/{{ ova_name }}.tmp"
+  file: path="{{ path }}" state=touch
+  register: ova_file
+  when: validate_only is not defined
 
--- a/roles/ovirt-ova-export-pre-pack/tasks/main.yml
+++ b/roles/ovirt-ova-export-pre-pack/tasks/main.yml
@@ -1,0 +1,31 @@
+---
+    - name: Examine target directory
+      stat:
+        path: "{{ target_directory }}"
+      register: target_directory_stats
+
+    - name: Fail if target directory does not exist
+      fail: msg="Target directory does not exist"
+      when: not target_directory_stats.stat.exists
+
+    - name: Fail if target directory is not a directory
+      fail: msg="Target directory is not a directory"
+      when: not target_directory_stats.stat.isdir
+
+    - name: Fail if target directory is not writeable
+      fail: msg="Target directory is not writeable"
+      when: not target_directory_stats.stat.writeable
+
+    - name: Removing the temporary file 
+      vars:
+        path: "{{ target_directory }}/{{ ova_name }}.tmp"
+      file: path="{{ path }}" state=absent
+      when: validate_only is not defined
+
+    - name: Prepare temporary path for the OVA file
+      vars:
+        path: "{{ target_directory }}/{{ ova_name }}.tmp"
+      file: path="{{ path }}" state=touch
+      register: ova_file
+      when: validate_only is not defined
+
--- a/roles/ovirt-ova-pack/files/pack_ova.py
+++ b/roles/ovirt-ova-pack/files/pack_ova.py
@@ -23,7 +23,7 @@ def padToBlockSize(file, size):
 
 ova_path = sys.argv[1]
 print "opening for write: %s" % ova_path
-ova_fd = os.open(ova_path, os.O_WRONLY | os.O_CREAT)
+ova_fd = os.open(ova_path, os.O_WRONLY)
 ova_file = io.FileIO(ova_fd, "w")
 
 ovf = sys.argv[2]
--- a/roles/ovirt-ova-pack/files/pack_ova.py
+++ b/roles/ovirt-ova-pack/files/pack_ova.py
@@ -23,39 +23,51 @@ def pad_to_block_size(file, size):
     if remainder > 0:
         file.write(NUL * (BLOCKSIZE - remainder))
 
-ova_path = sys.argv[1]
-print ("opening for write: %s" % ova_path)
-ova_fd = os.open(ova_path, os.O_WRONLY)
-ova_file = io.FileIO(ova_fd, "w")
 
-ovf = sys.argv[2]
-print ("writing ovf: %s" % ovf)
-ovf_size = len(ovf.encode('utf-8'))
-tar_info = create_tar_info("ovf", ovf_size)
-ova_file.write(tar_info.tobuf())
-ova_file.write(ovf)
-pad_to_block_size(ova_file, ovf_size)
+def write_ovf(ova_file, ovf):
+    print ("writing ovf: %s" % ovf)
+    ovf_size = len(ovf.encode('utf-8'))
+    tar_info = create_tar_info("ovf", ovf_size)
+    ova_file.write(tar_info.tobuf())
+    ova_file.write(ovf)
+    pad_to_block_size(ova_file, ovf_size)
 
-for disk_info in sys.argv[3:]:
-    # disk_info is of the following structure: <full path>::<size in bytes>
-    idx = disk_info.index('::')
-    disk_path = disk_info[:idx]
-    disk_size = int(disk_info[idx+2:])
+
+def write_disk(ova_file, disk_path, disk_size):
     print ("writing disk: path=%s size=%d" % (disk_path, disk_size))
     disk_name = os.path.basename(disk_path)
     tar_info = create_tar_info(disk_name, disk_size)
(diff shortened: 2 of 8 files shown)


ahadas:
@machacekondra , @mwperina , hi, some background:
This PR introduces a role for packing an OVA.
The general flow is:
1. the user selects a VM and a destination (host+path) to generate the OVA in
2. the engine creates a temporary disks (snapshots are collapsed, format is changed to qcow, later - having compression)
3. ansible is invoked to generate an OVA - it gets the OVF from the engine as well as the paths (+sizes) of the disks and creates the OVA in the specified destination

I would like to reuse this role in two contexts:
1. the user specifies the destination of the OVA - we will invoke that role with a destination that includes a predefined postfix (e.g., /tmp/vm.ova.tmp) and at the end of the process rename the file to remove that prefix (e.g., /tmp/vm.ova).
2. the user asks to download the OVA - we will write the OVA to an image created by the engine and later the engine downloads this image to the client

Ondra and I discussed this and thought about not having a role, but since I have a common task for these two flows I created one and I think about having two different playbooks for the two flows.

Please let me know if I'm in the right direction. Note that the python script may change, no need to review each detail there at this point.

Thanks!

roles/ovirt-ova/files/pack_ova.py
```diff
@@ -0,0 +1,53 @@
...
+    if remainder > 0:
+        tar.write(NUL * (BLOCKSIZE - remainder))
+
+# writing two null blocks at the end of the file
+empty = NUL * 512
+tar.write(empty)
+tar.write(empty)
+
+tar.close()
+
```
machacekondra: Why is there a need for this script? The `tar` utility is not enough in this case?

machacekondra (commented):
I am ok that it's a role, but we can also parametrized playbook, using variables which we send from engine. But currently we don't have any place in ovirt-engine where to store standalone scripts, so currently we have to go this way.

roles/ovirt-ova/files/pack_ova.py
ahadas: Unfortunately not. Imagine that the disk resides in a block storage device, invoking the tar with that disk as an input will result in an error that it is not a file. So we need to use something like FileIO

ahadas:
@machacekondra , @mwperina this PR is now ready for your review, thanks!

roles/ovirt-ova-export-pre-pack/tasks/main.yml
```diff
@@ -0,0 +1,31 @@
+---
+    - name: Examine target directory
```
machacekondra: Any reason to use indent here? I think it's OK, to be at index 0.

Why note do all those checks in the script? If the script is already there, it will be more effecient.

roles/ovirt-ova-pack/files/pack_ova.py
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/python
+
+import os
```
machacekondra: At the end of the day, this file will be stored in ovirt-engine, where we run pep8 checks, can you please re-sort alphabetically the imports?

roles/ovirt-ova-pack/files/pack_ova.py
```diff
@@ -0,0 +1,59 @@
...
+import io
+import tarfile
+import sys
+import time
+
+BLOCKSIZE = 512
+NUL = "\0"
+buf = bytearray(4096)
```
machacekondra, line 11: Can you please run pep8 on this file and fix all violations?

roles/ovirt-ova-pack/files/pack_ova.py
```diff
@@ -0,0 +1,59 @@
...
+
+def createTarInfo(name, size):
```
machacekondra: Can you please use snake case, instead of camel case?

roles/ovirt-ova-pack/files/pack_ova.py
```diff
@@ -0,0 +1,59 @@
...
+    info.mtime = time.time()
+    return info
+
+def padToBlockSize(file, size):
+    remainder = size % BLOCKSIZE
+    if remainder > 0:
+        file.write(NUL * (BLOCKSIZE - remainder))
+
+ova_path = sys.argv[1]
+print "opening for write: %s" % ova_path
```
machacekondra: Can we support both Python3 and Python2 please?

roles/ovirt-ova-pack/files/pack_ova.py
```diff
@@ -0,0 +1,59 @@
...
+ova_fd = os.open(ova_path, os.O_WRONLY)
```
machacekondra: Can this be in try-finally so we ensure we close it in case of failure?

roles/ovirt-ova-pack/files/pack_ova.py
```diff
@@ -0,0 +1,59 @@
...
+ova_file = io.FileIO(ova_fd, "w")
```
machacekondra: What's the use case when the target file exists?

roles/ovirt-ova-pack/files/pack_ova.py
```diff
@@ -0,0 +1,59 @@
...
+
+ovf = sys.argv[2]
```
machacekondra: At the script call in playbook, you don't require this parameter to be defined, so here it can fail.

roles/ovirt-ova-pack/tasks/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+  - name: Run packing script
+    script: pack_ova.py "{{ ova_file.dest }}" "{{ ovf }}" "{{ disks }}"
```
machacekondra: Can you please document somwhere how to run it? Some example playbook? And also what every variables means?

roles/ovirt-ova-pack/files/pack_ova.py
```diff
@@ -0,0 +1,59 @@
...
+print "writing ovf: %s" % ovf
+ovf_size = len(ovf.encode('utf-8'))
+tar_info = createTarInfo("ovf", ovf_size)
+ova_file.write(tar_info.tobuf())
+ova_file.write(ovf)
+padToBlockSize(ova_file, ovf_size)
+
+for disk_info in sys.argv[3:]:
```
machacekondra: I think this will fail in case disks is empty.

roles/ovirt-ova-pack/files/pack_ova.py
```diff
@@ -0,0 +1,59 @@
...
+    idx = disk_info.index('::')
+    disk_path = disk_info[:idx]
+    disk_size = int(disk_info[idx+2:])
```
machacekondra: Can you please comment what this does? Or what's the format of the `disk_info`?

roles/ovirt-ova-pack/files/pack_ova.py
```diff
@@ -0,0 +1,59 @@
...
+    print 'writing disk: path=%s size=%d' % (disk_path, disk_size)
+    disk_name = os.path.basename(disk_path)
+    tar_info = createTarInfo(disk_name, disk_size)
+    ova_file.write(tar_info.tobuf())
+    disk_fd = os.open(disk_path, os.O_RDONLY)
```
machacekondra: I think you don't close `disk_fd`

roles/ovirt-ova-export-pre-pack/tasks/main.yml
ahadas: indentation - fixed

in the python script? because I call the playbook twice - once in the validation/can-do-action phase of the command in the action with validate_only='True' so only those checks will be executed (without the packing phase). if I move them to the script, I'll need to pass that parameter to the script. I believe that in some phase we will be required to report those errors to the client so it will introduce another complexity - passing them from the python script to ansible

roles/ovirt-ova-pack/files/pack_ova.py
ahadas: done

roles/ovirt-ova-pack/files/pack_ova.py
ahadas: done

roles/ovirt-ova-pack/files/pack_ova.py
ahadas: done

roles/ovirt-ova-pack/files/pack_ova.py
ahadas: done

roles/ovirt-ova-pack/files/pack_ova.py
ahadas: that is intended for quick implementation for download, i.e., the user is able to download the ova to any destination by leveraging download-image. in the quick implementation, we create the target image in one of the storage domains in advance, and the script is supposed to dump the ova's content to that image which will then be downloaded

roles/ovirt-ova-pack/files/pack_ova.py
ahadas: how can enforce its definition? note that while this playbook is only used by the engine, these kind of validations are less important since they are done on the engine side

roles/ovirt-ova-pack/files/pack_ova.py
ahadas: added a comment that describes this

ahadas (commented):
not all comments are addressed yet

ahadas:
@machacekondra , I believe all comments are addressed now

ovirt-infra:
Some tests failed

roles/ovirt-ova-export-pre-pack/tasks/main.yml
```diff
@@ -0,0 +1,31 @@
+---
```
machacekondra, line 1: I think I've accidentally forgot to sent the most important comment.
I guess you care about performance here, so I think the better option for you would be to do those checks in the python script and remove both `ovirt-ova-export-pre-pack` `ovirt-ova-export-post-pack`. I think you can do this in python script. It will be more performant, as all of those tasks must package the python script(ansible module) and tranfer it via SSH to host, so you can safe 4 file transfer to host, if you will do those checks in the script.

ahadas: @machacekondra , I can put move the validations to the python script.
But let's say that we or the storage guys will next add the ability to export the VM to an OVA that is generated inside a pre-made disk image. In that case, we don't need to execute these validations and we don't need to rename the file. So if I move the content of both pre-pack and post-pack inside the python script, we would need different ansible role with a different python script that shares most of the code of this one. Would it be acceptable to merge such a thing later on? I just thought that ansible roles were meant to be reusable in different playbooks and so this design made sense to me, but I see your point regarding the performance.

machacekondra: Ah, OK, I understand. I just wanted to point out, that it could be more efficient. But I understand why you did it like this, definitely better when the script could be reused.

ahadas: @machacekondra , maybe you can help me with an issue I'm having - I posted a patch that adds the playbook and the roles to the engine's repository (https://gerrit.ovirt.org/#/c/80616/, shall we continue the review on gerrit btw?) and then the execution of the playbook fails with an error which says that the log file doesn't exist (the same code worked as part of this PR), so I commented out the log part and then it worked again (https://gerrit.ovirt.org/#/c/82526/40/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/CreateOvaCommand.java, line 122-125), did you encounter such a issue while moving other playbooks?

ahadas: @machacekondra actually there is a difference between the code on gerrit and the one here, I renamed the log directory (pack-ova -> create_ova) but I verified that this directory exists with the right permissions...

machacekondra: @ahadas Hmm, I didn't have such issue. If you commented out the log part, could you then see the log file in `var/log/ansible/ansible-{date}-{hostanme}.log` ? If yes, maybe there is some strange chars in the path.. can you share the log of ovirt-engine ?

And yes, if you can move the development to gerrit it would be better, as we already moved all the playbooks there, so if possible you can close it here. Thanks!

ahadas: @machacekondra , I found the issue - the log's directory was named create_ova and in the code I wrote create-ova... thanks :) so closing this PR, and we'll continue on gerrit.

PR #134, opened 2017-11-16
Add new ovirt-host-deploy-ovirt-metrics role and use it in ovirt-host…

sradco:
…-deploy role

* Add ovirt-host-deploy-ovirt-metrics

This role deploys oVirt metrics. It installs and configures
collectd and fluentd to send metrics and logs to a central
metrics store, if configured.

* Use new ovirt-host-deploy-facts role in host deploy role

This patch use ovirt-host-deploy-ovirt-metrics role in
ovirt-host-deploy-firewalld role.

This patch is not verified yet. @machacekondra @rgolangh


sradco:
This fixes https://bugzilla.redhat.com/show_bug.cgi?id=1494947

roles/ovirt-host-deploy-ovirt-metrics/README.md
```diff
@@ -0,0 +1,35 @@
+oVirt host deploy - oVirt metrics
```
machacekondra, line 1: I think we don't need it if there is no variables to document.

roles/ovirt-host-deploy-ovirt-metrics/tasks/main.yml
```diff
@@ -0,0 +1,4 @@
+---
```
machacekondra, line 1: Can you please create patch to oVirt engine, on top of: https://gerrit.ovirt.org/#/c/83870/ ?

roles/ovirt-host-deploy-ovirt-metrics/meta/main.yml
```diff
@@ -0,0 +1,21 @@
+galaxy_info:
```
machacekondra, line 1: We decided to store those roles internally and don't publish them on galaxy. So please remove this file.

roles/ovirt-host-deploy-ovirt-metrics/tasks/main.yml
```diff
@@ -0,0 +1,4 @@
...
+- name: Run oVirt metrics to configure metrics and logs collection on the host
+  command: /usr/share/ovirt-engine-metrics/setup/ansible/configure_ovirt_machines_for_metrics.sh --scope={{ inventory_hostname }}
```
machacekondra: Where is this file located on the host or engine?

roles/ovirt-host-deploy-ovirt-metrics/meta/main.yml
machacekondra: Ah, so I see you want them in galaxy based on:

 https://bugzilla.redhat.com/show_bug.cgi?id=1494947

Hmm, so we need to properly document it, first.

sradco: @rgolangh I don't think we need it in Galaxy. Roy can you please update here if we do and if it is not necessary, please update bugzilla. Thanks.

roles/ovirt-host-deploy-ovirt-metrics/tasks/main.yml
sradco: On the engine machine

machacekondra: Ok, then I think this will work good. The only issue could be dev-env, people need to make sure to have metrics installed always.

sradco: Actually they don't. If a metric store is configured then it will configure the host. Otherwise, it will show a message and end.

machacekondra: Well, this task will fail, because the command is not found, and so the host install will report `install_failed` status because ansible-playbook command will return return code `2`.

roles/ovirt-host-deploy-ovirt-metrics/README.md
sradco: Don't need the entire file? @machacekondra

ovirt-infra:
Some tests failed

roles/ovirt-host-deploy-ovirt-metrics/meta/main.yml
sradco: I talked to Roy and Galaxy is not required

machacekondra:
I am closing this one, as it was moved to: https://gerrit.ovirt.org/#/c/84331/

PR #141, opened 2017-12-01
Rename to ovirt-ansible-roles-all

mwperina:
Renames to ovirt-ansible-roles-all to prevent upgrade issues from older
releases which were not split to multiple RPMs

Signed-off-by: Martin Perina <mperina@redhat.com>


machacekondra:
ci build please

PR #146, opened 2018-03-07
automation: Fix syntax of 'release_branches'

ifireball:
'release_branches' were not properly specified, so builds were not
properly pushed into the 4.2 change-queue.

### DIFF
--- a/automation.yaml
+++ b/automation.yaml
@@ -3,6 +3,5 @@ distros:
   - fc26
   - el7
 release_branches:
-  master: ovirt-master
-  master: ovirt-4.2
+  master: [ "ovirt-master", "ovirt-4.2" ]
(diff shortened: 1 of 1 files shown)


ifireball:
@sandrobonazzola , @dron1 please note, we need this merged to properly release builds to 4.2

machacekondra:
This issue is also in other ansible repos, I will fix them.

eedri:
@ifireball is there a way the DSL parser can detect it and fail the patch if the syntax isn't correct?

ifireball:
@eedri no, there is no way to detect this since this is essentially specifying the same key twice in a dictionary, the YAML parser just picks one at random and the code has no way to tell that  this had happened.

PR #180, opened 2019-11-21
ovirt-ansible: bump ansible to 2.9

mnecas:
Changes:
- ansibel verson >= 2.9.0
- build version to 1.2.0
- and Requires ovirt roles >= 1.2.0


mwperina (commented):
Looks good to me, but we need to wait till all dependent roles are released

mwperina (approved):

PR #183, opened 2019-12-05
Update dependencies

mnecas:
Issue was when installing ovirt ansible roles it would install latest versions and not those with 1.1.Z.
@mwperina @machacekondra


mnecas:
ci build please

kobihk (approved):
verified with our 4.2 builder

machacekondra (approved):

